PR #27, opened 2018-11-28
add support for Tensorflow Keras API

c433li:
not unit tested


ririnicolae:
@c433li Thank you for submitting a pull request for ART! It is a good idea to provide support for TF Keras. The contribution introduces a high percentage of code redundancy, as your TF Keras support code is almost identical to the wrapper we already have for Keras models. As the logic is the same and most differences only come from imports, I would suggest providing the features by editing the current Keras support to add conditional imports. This would allow to import Keras from TF, or Keras directly, depending on what installs are detected in the user's environment. This would allow us to easily maintain this feature over time.

PR #78, opened 2019-05-24
Remove redundant parameter in adversarial patch

### LINKED ISSUE #77: Adversarial Patch size problem
By running 
`adv_patch = AdversarialPatch(classifier,patch_shape=(32, 32, 3), max_iter = 100)`
it turns out that patch_shape must have the same dimension as the input image. But in theory a patch could be smaller. Going into the source code the problem is by summing the image with the patch that is not done properly

ririnicolae:
# Description
The parameter `patch_shape` in the adversarial patch attack seems redundant, as the patch can only be created at the input size of the classifier. The patch is later rescaled to other dimensions in `apply_patch`. Having the parameter is misleading, as it creates the impression that the user can choose the initial size of the patch. I have thus removed it and changed the code to use the input size of the classifier instead. I have also updated the example notebook. This issue was brought forward by a user, see conversation in issue #77.

Fixes #77

## Type of change

- [ ] Bug fix (non-breaking)

# Testing
- [X] Unit tests

# Checklist

### DIFF
--- a/art/attacks/adversarial_patch.py
+++ b/art/attacks/adversarial_patch.py
@@ -35,10 +35,10 @@ class AdversarialPatch(Attack):
     """
 
     attack_params = Attack.attack_params + ["target", "rotation_max", "scale_min", "scale_max", "learning_rate",
-                                            "max_iter", "patch_shape", "batch_size", "clip_patch"]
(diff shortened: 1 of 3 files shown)


beat-buesser:
@ririnicolae Thank you for resolving this issue.

Issue #59, opened 2019-04-29
Create example notebook investigating the performance of boundary attack


ririnicolae:

ririnicolae:
Fixed in #79.

PR #79, opened 2019-05-27
Boundary

minhitbk:
Fixed issues in boundary attack and created a working notebook for it.


art/attacks/boundary.py
```diff
@@ -145,60 +162,59 @@ def _attack(self, initial_sample, original_sample, target, initial_delta, initia
...
         x_adv = initial_sample
-        delta = initial_delta
-        epsilon = initial_epsilon
+        self.curr_delta = initial_delta
+        self.curr_epsilon = initial_epsilon
+        clip_min, clip_max = self.classifier.clip_values
 
         # Main loop to wander around the boundary
-        for _ in range(self.max_iter):
+        for m in range(self.max_iter):
```
beat-buesser: Do we need to name the variable `m`? It seems it is not used inside of the loop.

art/attacks/boundary.py
```diff
@@ -63,67 +65,83 @@ def __init__(self, classifier, targeted=False, delta=0.01, epsilon=0.01, step_ad
...
+        else:
+            init_preds = [None] * len(x)
+            x_adv_init = [None] * len(x)
+
         # Assert that, if attack is targeted, y is provided
         if self.targeted and y is None:
             raise ValueError('Target labels `y` need to be provided for a targeted attack.')
 
         # Some initial setups
-        x_adv = x.astype(NUMPY_DTYPE)
```
beat-buesser: Would it be possible to use the previous `astype(NUMPY_TYPE)` here instead of `copy()`?
Both methods return a new copy of the array while the `astype` is consistent with all the other attack implementations and makes use if `NUMPY_DTYPE`.

art/attacks/boundary.py
```diff
@@ -63,67 +65,83 @@ def __init__(self, classifier, targeted=False, delta=0.01, epsilon=0.01, step_ad
...
+        x_adv = x.copy()
         if y is not None:
             y = np.argmax(y, axis=1)
-        preds = np.argmax(self.classifier.predict(x), axis=1)
 
         # Generate the adversarial samples
         for ind, val in enumerate(x_adv):
             if self.targeted:
-                x_adv[ind] = self._perturb(x=val, y_p=preds[ind], y=y[ind])
+                x_ = self._perturb(x=val, y=y[ind], y_p=preds[ind], init_pred=init_preds[ind], adv_init=x_adv_init[ind])
```
beat-buesser: Would it be possible to replace `x_` with `x_adv[ind]` and remove line 113?

art/attacks/boundary.py
```diff
@@ -63,67 +65,83 @@ def __init__(self, classifier, targeted=False, delta=0.01, epsilon=0.01, step_ad
...
             else:
-                x_adv[ind] = self._perturb(x=val, y_p=preds[ind])
```
beat-buesser: Would it be possible to replace `x_` with `x_adv[ind]` and remove line 113?

beat-buesser:
Fixes #59

moohax:
@beat-buesser II'm curious about this change, where targeted was set from `False` to `True`. Was there a particular reason for it? 

```python 
    def __init__(self, classifier, targeted=False, delta=0.01, epsilon=0.01, step_adapt=0.9, max_iter=100,
                 sample_size=20, init_size=100):

    def __init__(self, classifier, targeted=True, delta=0.01, epsilon=0.01, step_adapt=0.667, max_iter=5000,
                 num_trial=25, sample_size=20, init_size=100):
```

PR #93, opened 2019-06-18
Improve compatibility between preprocessing defences and standardisation

### LINKED ISSUE #84: Compatibility of defence JpegCompression with classifier's preprocessing (standardisation)
- In classifer.predict we are applying the classifier's preprocessing before applying the defences.
- Most of the common classifier preprocessing values for standardisation result in pixel values outside of [0, 1].
- The defence `JpegCompression` expects inputs `x` to be in the range [0, 1] and clips inputs `x` to this range [0, 1].
- This can very likely result in unexpected behaviour of the classifier without any warnings.

Are there any reasons against reversing the order to first applying the defences and then applying the preprocessing (standardisation) in classifier.predict()?
For example are there other defences that need to run after preprocessing (standardisation)?

- As alternative we could use the classifier's clip_values in JpegCompression to adapt the clipping for the compression step instead of just assuming [0, 1] for `x`.
- We could reverse the standardisation in `JpegCompression` and reapply it after the compression step, but this would result in extra computing efforts.
- At least we should print a warning if `JpegCompression` clips values in the input `x` that are outside of the expected range.

**Additional context**
The behaviour of `JpegCompression` is documented in its docstring.

beat-buesser:
# Description

This PR changes the order of the preprocessing steps of applying defences and standardisation/normalisation. Up to now the classifiers first applied standardisation followed by defences. This can result in unexpected negative and/or unbounded input values for the defences. With this PR the defences will be applied first followed by standardisation, that way we can feed to unprocessed input image values to the classifier and use the same and comparable defence parameters across classifiers with different standardisation/normalisation parameters.

Fixes #84 

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:
- OS
- Python version

# Checklist

### DIFF
--- a/art/classifiers/classifier.py
+++ b/art/classifiers/classifier.py
@@ -278,11 +278,33 @@ class Classifier(ABC):
         raise NotImplementedError
 
     def _apply_preprocessing(self, x, y, fit):
+        """
+        Apply all preprocessing steps of the classifier on inputs `(x, y)`.
+
+        :param x: Input data, where first dimension is the batch size.
+        :type x: `np.ndarray`
+        :param y: Labels for input data, where first dimension is the batch size.
+        :type y: `np.ndarray`
+        :param fit: `True` if the defences are applied during training.
+        :return: Value of the data after applying the defences.
+        :rtype: `np.ndarray`
+        """
         x_preproc = self._apply_preprocessing_normalization(x)
         x_defences, y_defences = self._apply_preprocessing_defences(x_preproc, y, fit=fit)
         return x_defences, y_defences, x_preproc
 
     def _apply_preprocessing_gradient(self, x, grads):
+        """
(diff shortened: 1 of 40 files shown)


beat-buesser:
I need to fix some unit tests, see #95, before the reviews can start.

beat-buesser:
Fixes #95

minhitbk:
Changes look fine to me, its ready for merging.

beat-buesser:
@ririnicolae Thank you for your initial feedback.

The change in the design of the defences originated in the idea to treat all attributes in the same way.

I don't think this puts an additional burden on the users because they will still be able to use the default values for the corresponding arguments and don't need to configure them.

I agree that that all variables should now be public, to remove the property for `apply_fit` and `apply_predict`, and to make it consistent among all defences. I'll push an update soon.

Issue #75, opened 2019-05-22
The loss_gradient of PyTorch should be carried on logits instead of softmax outputs


houxianxu:
**Describe the bug**
The loss calculated in [loss_gradient](https://github.com/IBM/adversarial-robustness-toolbox/blob/267a6e6ec4d490e52c81986e91f933cea5670e15/art/classifiers/pytorch.py#L331) is based on the model_outputs[-1] and the corresponding label. However, the [model_outputs[-1]](https://github.com/IBM/adversarial-robustness-toolbox/blob/267a6e6ec4d490e52c81986e91f933cea5670e15/art/classifiers/pytorch.py#L560)  in Pytorch Wrapper is the softmax outputs. Should model_ouputs[0] be used if we use cross_entropy loss like in the [test of Pytorch ](https://github.com/IBM/adversarial-robustness-toolbox/blob/c65512b6ce64976a68fb2d9a19e132b5f6dddd0a/tests/classifiers/test_pytorch.py#L62)?

For convenient, related codes are shown below
1. The code to compute model_outputs:
```Python
result = []
if isinstance(self._model, nn.Sequential):
    for _, module_ in self._model._modules.items():
        x = module_(x)
        result.append(x)
elif isinstance(self._model, nn.Module):
    x = self._model(x)
    result.append(x)
else:
    raise TypeError("The input model must inherit from `nn.Module`.")
output_layer = nn.functional.softmax(x, dim=1)
result.append(output_layer)
return result
```

2. The code to compute the loss:
```python
# Compute the gradient and return
model_outputs = self._model(inputs_t)
loss = self._loss(model_outputs[-1], labels_t)
```

minhitbk:
Hi @houxianxu :  Thanks alot for the issue you have opened. We have identified it and will fix it in the next few days.

ririnicolae:
This issue will be solved later this week by the solution for epic #50.

ririnicolae:
Fixed in #119 (on `dev` branch). Loss function can now be specified by the user for `PyTorch` classifier and will be applied directly on the output of the model.

PR #119, opened 2019-07-20
Generalize classifiers from TF / Keras /  PyTorch / MXNet to other types of models than neural networks

### LINKED ISSUE #50: Adapt ART to other types of models than neural networks
This mostly implies changing the way different backend implementations handle model outputs. In order to be effective, most attacks need to be applied at logits level. Some of the new models types ART will support do not have logits. The following describes the proposed solution for this problem:

- All backend `__init__` functions would have a flag specifying if the output of the given model represents logits. When such a model is created, attacks are applied on the regular model output, as was done previously.
- If the outputs are not logits, we can check if the sum of the outputs on a fake input is 1. In that case, we have detected probabilities as outputs and will try to estimate the logits from them. These would not be the real logits of the model, but would perform numerical scaling that makes attacks efficient.
- If the outputs are not logits and don't sum to 1 either, this might be a different type of model, and attacks will be applied directly on its outputs.
- Another flag could indicate that the outputs of the model are probabilities, but the user explicitly requests that the previous layer (logits) should be used for attacks. This option is specific to the frameworks that support the notion of layer. Moreover, finding the correct logits would not be guaranteed.
- The loss function for a given model will also be part of the input provided by the user. The loss will be applied directly on the output of the model.

In practice, the implications are that predictions / gradient computations will not be available at both logits and probabilities level, as it is currently the case. This would attract a change in the `Classifier` API for the respective functions.

ririnicolae:
# Description

- These changes impact TF / Keras / PyTorch / MXNet classifiers.
- They now all accept custom / general loss functions taken as input parameters.
- Parameter `logits=True/False` has been removed from `Classifier` API in functions `predict` and `class_gradient`. The predictions / gradients are now computed at the output point of the model.
- All attacks and other methods have been updated to the changes of the interface.
- Examples / notebooks have been updated.
- Classifiers' docstrings have been updated to inform users that logit outputs are more effective for attack success.
- One issue was discovered during this extension: because model predictions can be negative, KL divergence computation in VAT sometimes yields `nan`s. I will open a separate issue for this.
- See also #50 for additional comments.

Fixes #50, #75

## Type of change

- [ ] Bug fix (non-breaking)
- [X] New feature (non-breaking)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Testing

- [X] Unit tests

# Checklist


ririnicolae:
@beat-buesser Thanks for merging! To answer your questions:

- Yes, the commented test in VAT is the one that fails as described in #120.
- When the `logits` parameter was part of the `class_gradient` function signature, the `DetectorClassifier` had 8 cases to test (4 with `logits=True` and 4 with `False`). Without the parameter, only 4 cases remain to be tests, thus the change.

PR #129, opened 2019-08-04
Example of attacking MNIST using pytorch using FGSM

### LINKED ISSUE #128: Pytorch examples
**Description**

It is extremely difficult to start using the toolbox just by looking at the docs or the source code . The examples given in the *examples* folder seem to be extremely useful. However, all the files in the *examples* folder are using Keras or Tensorflow. Support for other popular frameworks such as Pytorch would be extremely useful.

ksivaman:
Signed-off-by: ksivaman <ksivaman@purdue.edu>

# Description

**Description**
An example of using ART using Pytorch. I previously pointed out a lack of ART examples in pytorch in #128  . This attack uses the untargeted attack (FGSM) on a pytorch classifier trained on the MNIST dataset. The results obtained are given below. I feel the issue #128 can be closed after this addition.

Fixes #128 

This is an addition of a simple example and does not break any existing features/functions. I have tested the function and have obtained the following results. The script can be run independently and this is an end to end example.  

Classification accuracy on MNIST dataset before attack: 97.26%

Classification accuracy after attack using fast-gradient-sign-method: 22.06%

### DIFF
--- a/examples/fgsm_pytorch_mnist.py
+++ b/examples/fgsm_pytorch_mnist.py
@@ -1,0 +1,61 @@
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+import numpy as np
+
+from art.attacks.fast_gradient import FastGradientMethod
+from art.classifiers.pytorch import PyTorchClassifier
+from art.utils import load_mnist
+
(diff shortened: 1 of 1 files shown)


beat-buesser:
@ksivaman Thank you for your first contribution to ART!

Issue #138, opened 2019-08-15
Variable prediction output when learning parameter is set in keras classifiers


robo-sq:
Variable predictions when learning parameter is set in keras classifiers, k.set_learning_phase(1).
Only affects notebooks: mnist_poisoning_demo and randomized_smoothing_mnist and examples: cifar_adversarial_training.py, mnist_cnn_features_level_fgsm.py, mnist_cnn_fgsm.py, mnist_poison_detection.py.

PR #139, opened 2019-08-15
removing set learning parameter in noteooks and examples using keras classifiers

robo-sq:
fixing issue #138 

removing set learning paramter from notebooks and example files.


beat-buesser:
@robo-sq Thank you very much for noticing and fixing it!

Issue #157, opened 2019-09-09
Investigate vector support issue with VAT


minhitbk:
**Describe the bug**
The test_iris_pt and the test_iris_tf get the same error in the unittest of VAT so
they was commented out temporarily. The problem is that the attack generates  "nan" in both cases so that could be a problem with the attack / or it simply doen't support vector.

minhitbk:
This issue is related to issue #120 as well.

PR #159, opened 2019-09-09
Fix invalid computations in VAT

### LINKED ISSUE #120: Invalid values in KL divergence computation of VAT
**Describe the bug**
This issue was discovered during the development for #75. Because the model predictions / logits can be negative, the computation of the KL divergence in virtual adversarial method can yield invalid values (`nan` and `inf`), preventing the method from producing meaningful results.

 - OS: Ubuntu
 - Python version: 3.6
 - ART version or commit number: 3d618e3

ririnicolae:
# Description

Fix issue with numerical computation in VAT. As described in the original issue (#120), one of the root causes was related to invalid values encountered in the computation of KL divergence. This was coming from negative values predicted in classifiers. Moreover, an additional root cause was present in the computation of L_2 data normalization in VAT. This PR fixes both issues. As a consequence, VAT can be used on both images and feature vectors without restrictions and warnings.

Fixes #120, #157

## Type of change

- [X] Bug fix (non-breaking)

# Testing

- [X] Existing unit tests
- [X] Uncommented unit tests that were previously failing

**Test Configuration**:
- OS: Ubuntu
- Python version: 3.6

# Checklist


minhitbk:
I am a bit confused on the _rescale function. I had a quick look at the paper and the original code of VAT of the authors, I dont see that rescale concept (or maybe I missed it). I think VAT is designed to work with probability output, so users need to guarantee that their model should produce probability. In this case, we don't need to rescale or change the current implementation of VAT (except for the small bug in the normalize function). Instead, we only need to raise error to users if they input a non-probability output model, and fix bugs in the unittests.

In case the concept of rescale is available somewhere in VAT but I missed, then the code looks well to me and ready to merge.

ririnicolae:
@minhitbk You are right, the `_rescale` function is not part of the original VAT paper. As you mentioned, the method was conceived to only be applied on top of softmax. However, ART now handles many more types of models than DNNs with softmax. Introducing the `_rescale` function is an extension that allows to generalize to these models. I thought this would be a good idea, considering we have adapted other methods in the past for better performance or extended use cases.

Issue #192, opened 2019-10-31
What changed with DeepFool in 1.0.0?


marcoancona:
**Describe the bug**
I have a simple Keras CNN with Softmax activation, trained on MNIST.
These was the result with ART 0.10.0:
![image](https://user-images.githubusercontent.com/2278117/67958518-9530ad80-fbf7-11e9-8474-ed12dcfdfa33.png)

And this with 1.0.1
![image](https://user-images.githubusercontent.com/2278117/67958209-18053880-fbf7-11e9-91bf-68c117f4113d.png)

As you can see, without any other change, on 1.0 the adversarial accuracy is nearly 0 but only because the input images are totally destroyed (not classifiable by a human). The result with ART 0.10 seems much more reasonable. 

What has changed? If I go through the changelog, I see that it might be related to the following point:

> Generalize the classifiers of TensorFlow, Keras, PyTorch, and MXNet by removing assumptions on their output (logits or probabilities). The Boolean parameter logits has been removed from Classifier API in methods predict and class_gradient. The predictions and gradients are now computed at the output of the model without any modifications. (#50, #75, #106, #115 

This also confused me because `use_logits` is still available on KerasClassifier so I thought the behavior would have been the same.

**System information (please complete the following information):**
 - Ubuntu 16.04
 - Python version: 3.6
 - ART version or commit number: 0.10.0 vs 1.0.1
 - TensorFlow 1.11 / Keras 2.2.4

beat-buesser:
Hi @marcoancona Thank you very much for using ART and raising this issue. We will look at it as soon as possible. In the meantime, would you have a script available to share with us that reproduces the described behaviour?

marcoancona:
I will try to build a minimal example.
One suspicious line is https://github.com/IBM/adversarial-robustness-toolbox/blob/9c6ebc6567bb1533e3048973b02fd146cc1a73cc/art/classifiers/keras.py#L162
Seems like `use_logits` is ignored with `categorical_crossentropy`. Why?

beat-buesser:
Hi @marcoancona 

Ok, is it correct that for logits as output the behaviour in both versions is as expected?

The example in this Stack Overflow post shows how to train a Keras model that predicts logits: https://stackoverflow.com/questions/47036409/keras-how-to-get-unnormalized-logits-instead-of-probabilities

Many attacks are much stronger on logits than on probabilities.

[...]

PR #212, opened 2019-11-12
Improve loss function handling in KerasClassifier

beat-buesser:
# Description

This pull request improves the handling of loss functions in `KerasClassifier` and enables the use of loss function generators which facilitates using models that predict logits. It updates the unittests for keras and tf.keras and tests each supported loss function type and implementation with probabilities and logits, where possible, and checks for identical results in `predict`, `loss_gradient` and `class_gradient`.

Fixes issues discovered as part of the discussion in #192

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


art/utils_test.py
```diff
@@ -210,38 +211,109 @@ def get_classifier_kr(loss_name='categorical_crossentropy'):
...
+                    return keras.backend.sparse_categorical_crossentropy(y_true, y_pred, from_logits=True)
+
+                loss = sparse_categorical_crossentropy
+            else:
+                loss = keras.backend.sparse_categorical_crossentropy
     elif loss_name == 'kullback_leibler_divergence':
-        loss = keras.losses.kullback_leibler_divergence
-    elif loss_name == 'cosine_proximity':
-        loss = keras.losses.cosine_proximity
+        if loss_type == 'label':
```
ririnicolae: Some of the if-elses around `loss_type` can probably be avoided by introducing a `dict` that maps the loss type to the action, then doing something like `loss = dict_loss_type[loss_type]`. The dict value could even be a function that does some of the checks required. This goes for all constructs of this type in the test models.

tests/classifiers/test_keras.py
```diff
@@ -397,12 +397,171 @@ def test_repr(self):
...
+                        loss_type.format(loss_name, loss_type))
+
+            _run_tests(loss_name, loss_type, y_test_pred_expected, class_gradient_probabilities_expected,
+                       loss_gradient_expected, _from_logits=False)
+
+        # # ================= #
+        # # cosine_similarity #
+        # # ================= #
+        #
+        # loss_name = 'cosine_similarity'
```
ririnicolae: Should this test stay commented?

tests/defences/test_spatial_smoothing.py
```diff
@@ -39,6 +39,7 @@ def test_ones(self):
 
         # Start to test
         for window_size in range(1, 20):
+            print('Window size: {}'.format(window_size))
```
ririnicolae: Why output the same info in logger and print?

tests/classifiers/test_keras.py
```diff
@@ -397,12 +397,171 @@ def test_repr(self):
...
-            classifier.loss_gradient(x=self.x_test, y=self.y_test)
+
+        # prediction and class_gradient should be independent of logits/probabilities and of loss function
+
+        y_test_pred_expected = np.asarray([7, 1, 1, 4, 4, 1, 4, 4, 4, 4, 4, 4, 4, 4, 1, 4, 4, 7, 8, 4, 4, 4, 4, 3, 4, 4,
+                                           7, 4, 4, 4, 7, 4, 3, 4, 7, 0, 7, 7, 1, 1, 7, 7, 4, 0, 1, 4, 4, 4, 4, 4, 4, 7,
+                                           4, 4, 4, 4, 4, 1, 4, 4, 7, 4, 4, 4, 4, 4, 4, 4, 3, 0, 7, 4, 0, 1, 7, 4, 4, 7,
+                                           4, 4, 4, 4, 4, 7, 4, 4, 4, 4, 4, 1, 4, 4, 4, 4, 4, 4, 4, 4, 4, 4])
+
+        class_gradient_probabilities_expected = np.asarray([0.0000000e+00, 0.0000000e+00, 0.0000000e+00, 2.3582461e-03,
```
ririnicolae: The recent tendency has been to assert exact values in ART tests, but these are generally not very stable in the long run and require a lot of maintenance. Depending on the machine architecture and versions of packages, these numbers are not necessarily reproducible. E.g., in my case, versions of ART that pass on Travis fail locally. Should we think about another way of testing effectively?

tests/defences/test_spatial_smoothing.py
beat-buesser: I agree, the print statement needs to be removed again, it got forgotten. I had added it for debugging because this test sometimes fails and the test logging for this test is not showing up in the Travis logs.

beat-buesser: Fixed in 614a65395b392fe959e50c267e6cc6517e4d22f5

tests/classifiers/test_keras.py
beat-buesser: I agree that recent versions of ART had too much variability in some of these new unittests, but I think we have removed or improved most of these unstable tests on `dev_1.1.0`. The unstable testes were mostly for classes containing algorithms with a lot of randomness and initially we thought that random seeds would fix it. I like this kind of tests in cases where the results are reproducible, because they run fast and capture changing behaviour quite well. But I'm also open to other types of testing too.

tests/classifiers/test_keras.py
beat-buesser: I will remove this loss function for now because its name changes between Keras/TF versions.

art/utils_test.py
beat-buesser: I agree and have tried to create a solution based on a dictionary but I feels like it becomes a quite complicated dictionary. I tried to improve the two functions by extending the docstrings and adding exception messages.

beat-buesser:
Hi @ririnicolae Thank you for your great review! Ok, I'll wait with merging until tomorrow.

PR #231, opened 2019-12-06
Implement Functionally Equivalent Extraction attack

### LINKED ISSUE #216: Implement Functionally Equivalent Extraction Attack
Implement Functionally Equivalent Extraction Attack for neural networks with 2 fully-connected layers with ReLU activation and logits output as described by Jagielski et al. 2019 (https://arxiv.org/abs/1909.01838):.

beat-buesser:
# Description

This pull request introduces the first extraction attack to ART. The Functionally Equivalent Extraction attack follows Jagielski et al., 2019 and is able to extract weights and biases of a neural network with two dense layers with ReLU activation at the first layer and logits output after the second layer.

Fixes #216 

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


beat-buesser:
@minhitbk Thank you very much for the detailed review:
> 1. Line 182: Why you don't use the _get_x function as in lines 180, 183, 184?

This line must have been forgotten, I have added `_get_x` in line 182

> 2. In the [ while not found_critical_point ] loop, the delta param is always divided by 2 everytime many critical_points are found, what happens if at some point when we divide delta by half, the range [t1, t2] doesn't contains any critical_point, so should we somewhere increase delta or move this range forward to cover at least 1 critical_point?

If there is not critical point between `t_1` and `t_2` then line 191 will break the while loop of line 173 which resets `delta` to `delta_0` and sets `t` to the current `t_2`, afterwards search continues again form the new `t`. 

> 3. I suggest that the class FunctionallyEquivalentExtraction should be extended from the class Attack for consistency in ART.

I agree, I have added the inheritance from `Attack`.

PR #232, opened 2019-12-06
implement copycat cnn attack

### LINKED ISSUE #229: Implement copycat cnn attack
Implement the attack Copycat CNN, https://arxiv.org/abs/1806.05476, in the Epic #191

minhitbk:
# Description
Implement the copycat cnn attack.

Fixes #229 

## Type of change

- [ ] Bug fix (non-breaking)
- [X] New feature (non-breaking)

# Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:
- OS
- Python version

# Checklist


art/attacks/extraction/copycat_cnn.py
```diff
@@ -0,0 +1,151 @@
...
+        """
+        Create a copycat cnn attack instance.
+
+        :param classifier: A victim classifier.
+        :type classifier: :class:`.Classifier`
+        :param batch_size: Size of batches.
+        :type batch_size: `int`
+        :param nb_epochs: Number of epochs to use for training.
+        :type nb_epochs: `int`
+        :param nb_stolen: Number of examples to be stolen.
```
beat-buesser: Shouldn't this and similar lines below be something like "the number of examples used for stealing the model"? This attack is not stealing the examples.

art/attacks/extraction/copycat_cnn.py
```diff
@@ -0,0 +1,151 @@
...
+
+        :param x: An array with the source input to the victim classifier.
+        :type x: `np.ndarray`
+        :param y: Target values (class labels) one-hot-encoded of shape (nb_samples, nb_classes) or indices of shape
+                  (nb_samples,). Not used in this attack.
+        :type y: `np.ndarray` or `None`
+        :param thieved_classifier: A thieved classifier to be stolen.
+        :type thieved_classifier: :class:`.Classifier`
+        :return: The stolen classifier.
+        :rtype: :class:`.Classifier`
```
beat-buesser: `generate` as defined in the abstract base class `Attack` defines the return type as an `np.ndarray` (of adversarial examples). Maybe we should define a new method that returns a classifier?

minhitbk: @beat-buesser I agree, should we define an abstract method in the Attack class, so that will create a standard for all model extraction attacks, I suggest the method name "extract"?

beat-buesser: Solved after discussion with @minhitbk with #250

art/attacks/extraction/copycat_cnn.py
imolloy: Or "Number of queries submitted to the model to steal it"

art/attacks/extraction/copycat_cnn.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Select data to attack
+        selected_x = self._select_data(x)
+
+        # Query the victim classifier
+        fake_labels = self._query_label(selected_x)
+
+        # Train the thieved classifier
+        thieved_classifier.fit(x=selected_x, y=fake_labels, batch_size=self.batch_size, nb_epochs=self.nb_epochs)
+
```
imolloy: Does it make sense to ensure `thieved_classifier` is a `Classifier` or just that it has a `fit` function?

art/attacks/extraction/copycat_cnn.py
```diff
@@ -0,0 +1,151 @@
...
+        Query the victim classifier.
+
+        :param x: An array with the source input to the victim classifier.
+        :type x: `np.ndarray`
+        :return: Target values (class labels) one-hot-encoded of shape (nb_samples, nb_classes).
+        :rtype: `np.ndarray`
+        """
+        labels = self.classifier.predict(x=x, batch_size=self.batch_size)
+        labels = np.argmax(labels, axis=1)
+        labels = to_categorical(labels=labels, nb_classes=self.classifier.nb_classes())
```
imolloy: Is this specific to the paper cited? I would rather see the option to return soft labels which makes some of these attacks more effective. Perhaps as an argument to the `__init__` or `generate`. The ability to handle the soft labels depends on the `thieved_classifier`.

art/attacks/extraction/copycat_cnn.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    attack_params = Attack.attack_params + ['batch_size', 'nb_epochs', 'nb_stolen']
+
+    def __init__(self, classifier, batch_size=1, nb_epochs=10, nb_stolen=1):
+        """
+        Create a copycat cnn attack instance.
+
+        :param classifier: A victim classifier.
+        :type classifier: :class:`.Classifier`
+        :param batch_size: Size of batches.
```
imolloy: Just realized we've overloaded batch size here to be both the number of samples to pass the model in the `_query_label` and the `fit`.

art/attacks/extraction/copycat_cnn.py
minhitbk: This is done.

art/attacks/extraction/copycat_cnn.py
minhitbk: This is done.

art/attacks/extraction/copycat_cnn.py
minhitbk: As discussed with @beat-buesser , we would like to keep the thieved_classifier an instance of the ART Classifier for 2 reasons: 1. If we just check for the availability of the fit function, we may not know the signature of that function for use, 2. ART supports a BlackClassifier that allows any type of classifier to be wrapped as an ART classifier.

art/attacks/extraction/copycat_cnn.py
minhitbk: Thanks @imolloy , this is a cool comment and I agree that we should have a separate batch_size for each of the classifiers: 1 for training the thieved_classifier and the other for querying the victim_classifier.

art/attacks/extraction/copycat_cnn.py
minhitbk: Thanks @imolloy , this is also a cool idea to keep a soft version of labels. However, as discussed with @beat-buesser , the thieved_classifier which is expected to be an ART classifier and therefore only accepts 1-hot labels, we temporarily keep this type of hard labels. We need to investigate more if changing ART classifier's fit function will cause any trouble for ART. See issue #251

PR #240, opened 2019-12-08
Enable top level package import

### LINKED ISSUE #235: Enable import of full ART package
It would be nice to be able to import the full ART toolkit using `import art`

However, the import structure does not allow this:
```
import art
wrapper = art.classifiers.KerasClassifier(model=dummy_model)
```
Returns `AttributeError: module 'art' has no attribute 'classifiers'`

See: 
https://colab.research.google.com/drive/1X3U0xm0uEEwWDM4njwwbl9aDK5Jk9OeO

I would be willing to submit a PR for this but wanted to check with the maintainers to see if there is a reason for this.

seanpmorgan:
# Description
Enable top level import. Happy to move the configs somewhere else if you'd prefer.

```
import art
art.classifiers.keras.KerasClassifier
```
* Import submodules from package init
* Move DATA_PATH and NUMPY_DTYPE to utils so that there is no circular dependency

Fixes #235 

- [ ] Bug fix (non-breaking)

# Testing
Can add a unit test if you would like. Other unit tests likely cover the changes though.

# Checklist


beat-buesser:
Hi @seanpmorgan Thank you for this pull request!

beat-buesser:
I have asked @ririnicolae for a review @ririnicolae is currently looking at the logging of ART and I also would like to check if this effects the work on logging.

art/utils.py
```diff
@@ -434,7 +473,7 @@ def load_batch(fpath):
         data = data.reshape(data.shape[0], 3, 32, 32)
         return data, labels
 
-    from art import DATA_PATH
+    from art.utils import DATA_PATH
```
beat-buesser: Is this import still required? It results in a LGTM warning on importing itself.

art/utils.py
```diff
@@ -477,7 +516,7 @@ def load_mnist(raw=False):
     :return: `(x_train, y_train), (x_test, y_test), min, max`
     :rtype: `(np.ndarray, np.ndarray), (np.ndarray, np.ndarray), float, float`
     """
-    from art import DATA_PATH
+    from art.utils import DATA_PATH
```
beat-buesser: Is this import still required? It results in a LGTM warning on importin itself.

art/utils.py
```diff
@@ -508,7 +547,7 @@ def load_stl():
     :rtype: `(np.ndarray, np.ndarray), (np.ndarray, np.ndarray), float, float`
     """
     from os.path import join
-    from art import DATA_PATH
+    from art.utils import DATA_PATH
```
beat-buesser: Is this import still required? It results in a LGTM warning on importin itself.
```suggestion
```

art/utils.py
```diff
@@ -553,7 +592,7 @@ def load_iris(raw=False, test_set=0.3):
     :return: Entire dataset and labels.
     :rtype: `(np.ndarray, np.ndarray), (np.ndarray, np.ndarray), float, float`
     """
-    from art import DATA_PATH, NUMPY_DTYPE
+    from art.utils import DATA_PATH, NUMPY_DTYPE
```
beat-buesser: Is this import still required? It results in a LGTM warning on importin itself.
```suggestion
```

art/utils.py
```diff
@@ -664,7 +703,7 @@ def get_file(filename, url, path=None, extract=False):
     :rtype: `str`
     """
     if path is None:
-        from art import DATA_PATH
+        from art.utils import DATA_PATH
```
beat-buesser: Is this import still required? It results in a LGTM warning on importin itself.
```suggestion
```

seanpmorgan:
@beat-buesser Thanks for the review. Apologies for the self imports, missed that and they've been removed.

I wouldn't import `art.utils` in `art/__init__` because it'll import all the globals defined in that file during `import art`.  Only the subpackages are imported here so that `art.supabakge.*` can be called.

seanpmorgan:
> Hi, @seanpmorgan ! Thank you for your contribution, I think it's indeed convenient to have shorthands for imports. However, moving the logic in `art/__init__.py` to `art/utils.py` has some unintended side effects:
> 
> * That code is boilerplate logic that needs to run when `art` is imported, however it is imported. This is not the case anymore when importing just `art`.
> * Moreover, in order for the logic that was moved to be executed, now developers need to import `art/utils.py` by hand in every new file. This is bound to introduce errors in the future, as there is no way to guarantee it.
> 
> I suggest that the boilerplate init logic returns to `art/__init__.py`. I understand that the change was introduced to avoid circular imports, but we should probably find another way around that problem. Let us know if we can help with that.

Hi @ririnicolae thanks for the review!

So currently the boiler plate `NUMPY_DTYPE` and `DATA_PATH` have to be imported in all submodules of ART that use it (see all of the files modified in this PR). This is true before and after this PR. If an end user wanted to use these global variables they would not be able to without `import art`; and as described in #235 that import pattern is pretty useless. 

```
(art) sean-morgan@ubuntu-spmorgan:~/code/adversarial-robustness-toolbox$ python
Python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 19:07:31) 
>>> from art.attacks.boundary import BoundaryAttack
>>> print(DATA_PATH)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'DATA_PATH' is not defined
```

I also want to be clear about the issue this is trying to solve. This isn't really "short-hand" imports but rather allowing a user to `import art` and then access all of the submodules beneath it (currently not allowed). 

I'm happy to move the boilerplate out of `utils` and I agree that's not the best place for it... but placing them in `art.__init__` will make it so that the top level init is unable to import submodules (circular import since the submodules have to import art -- an anti-pattern). 

As a side note I also think `DATA_PATH` and `NUMPY_DTYPE` are pretty dangerous global variable names. Perhaps using `ART_DATA_PATH` would be better in case an end-user decides to recreate those variables unknowingly.

ririnicolae:
@seanpmorgan Thanks for your detailed response to the review! I agree with all the points you have mentioned. Would it work for you to move the global vars to something like `art/config.py` that would only deal with the configuration file and these global values?

From my point of view, feel free to also add the prefix `ART_` to the var names, as you suggested. @beat-buesser What do you think?

beat-buesser:
@ririnicolae I think moving the boilerplate code to `art/config.py` makes sense. I also like the idea of prepending `ART_` to global variable names to avoid future naming conflicts.

@seanpmorgan I think you can go ahead and implement the discussed changes.

seanpmorgan:
Updated. 

**NOTE:** There is a side-effect of re-naming these variables. Because the config first tries to load from `~/.art/config.json` if the user had previously installed and written this config then imports will fail. Is there a way you would like to handle this issue?

art/config.py
```diff
@@ -0,0 +1,45 @@
+import json
+import logging
+import os
+
+import numpy as np
+
+logger = logging.getLogger(__name__)
+
+
+# -------------------------------------------------------------------------------------------- DEFUALT PACKAGE CONFIGS
```
beat-buesser:
```suggestion
# -------------------------------------------------------------------------------------------- DEFAULT PACKAGE CONFIGS
```

art/config.py
```diff
@@ -0,0 +1,45 @@
...
+_folder = os.path.join(_folder, '.art')
+
+_config_path = os.path.expanduser(os.path.join(_folder, 'config.json'))
+if os.path.exists(_config_path):
+    try:
+        with open(_config_path) as f:
+            _config = json.load(f)
+    except ValueError:
+        _config = {}
+
```
beat-buesser: Would it be possible to check here if `_config` contains a key `DATA_PATH` and if yes remove it and add the key `ART_DATA_PATH` with the value received for `DATA_PATH`  and overwrite config.json?

beat-buesser:
@seanpmorgan Thank you for the changes and the note on potential side-effects. What do you think about the proposed solution above?

seanpmorgan:
> @seanpmorgan Thank you for the changes and the note on potential side-effects. What do you think about the proposed solution above?

Sounds like a good solution to me. Updated.

beat-buesser:
@seanpmorgan Looks good to me. Thank you very much for this great contribution!

seanpmorgan:
> @seanpmorgan Thanks for the edits! Does it make more sense now to import `utils` in `__init__`? The circular dependency should be gone. In any case, PR looks good to me.

@ririnicolae I think we should be all set. Since the modules (utils/visualization/etc.) are within top-level `art/` and not under subpackages they should be able to be accessed with `import art`:

```
Python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 19:07:31) 
>>> import art
>>> art.classifiers.keras.KerasClassifier
<class 'art.classifiers.keras.KerasClassifier'>
>>> dir(art.utils)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_extract', 'absolute_import', 'check_and_transform_label_format', 'clip_and_round', 'compute_accuracy', 'compute_success', 'division', 'get_file', 'get_label_conf', 'get_labels_np_array', 'least_likely_class', 'load_cifar10', 'load_dataset', 'load_iris', 'load_mnist', 'load_stl', 'logger', 'logging', 'make_directory', 'master_seed', 'np', 'original_to_tanh', 'os', 'performance_diff', 'preprocess', 'print_function', 'projection', 'random_sphere', 'random_targets', 'second_most_likely_class', 'segment_by_class', 'tanh_to_original', 'to_categorical', 'unicode_literals']
>>> art.utils.segment_by_class
<function segment_by_class at 0x7fd37326fb70>
```

PR #282, opened 2020-02-04
Added Pixel and Threshold Attack

shashankkotyan:
# Description
Added implementation of Few-Pixel and Threshold Attack. 
Requires pycma (https://github.com/CMA-ES/pycma) library. 

Resolves #280  and #281 

## Type of change
- [x] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# Testing
Tested Pixel and Threshold on the CIFAR-10 dataset across Keras and TensorflowV2 platforms. 

- [x] Pixel Attack with DE optimisation on Keras Classifier
- [x] Threshold Attack with DE optimisation on Keras Classifier
- [x] Pixel Attack with CMA-ES optimisation on Keras Classifier
- [x] Threshold Attack with CMA-ES optimisation on Keras Classifier
- [x] Pixel Attack with DE optimisation on TesnorflowV2 Classifier
- [x] Threshold Attack with DE optimisation on TesnorflowV2 Classifier
- [x] Pixel Attack with CMA-ES optimisation on TesnorflowV2 Classifier
- [x] Threshold Attack with CMA-ES optimisation on TesnorflowV2 Classifier

**Test Configuration**:
- OS Ubuntu 18.04
- Python version 3.7
- ART version 1.1.0 on dev 1.2.0 branch
- TensorFlowV2 2.1.0
- Keras 2.2.3 with Tensorflow 1.13.1

# Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #287, opened 2020-02-08
Re-Added Pixel and Threshold Attack

### LINKED ISSUE #280: Request for addition of Few Pixel Attack (One Pixel Attack)
I would like to add an implementation of Few Pixel Attack (https://arxiv.org/pdf/1906.06026.pdf) which is a generalisation of One Pixel Attack (https://arxiv.org/pdf/1710.08864.pdf).

### LINKED ISSUE #281: Request for addition of Threshold Attack
I would like to add an implementation of Threshold Attack (https://arxiv.org/pdf/1906.06026.pdf).

shashankkotyan:
# Description
Added implementation of Few-Pixel and Threshold Attack. 
Requires pycma (https://github.com/CMA-ES/pycma) library. 

Resolves #280  and Resolves #281 

## Type of change
- [ ] Bug fix (non-breaking)
- [x] This change requires a documentation update

# Testing
Tested Pixel and Threshold on the CIFAR-10 dataset across Keras, TensorflowV1, TensorflowV2 and PyTorch platforms. 
- [x] Pixel Attack with DE optimisation
- [x] Threshold Attack with DE optimisation
- [x] Pixel Attack with CMA-ES optimisation
- [x] Threshold Attack with CMA-ES optimisation

**Test Configuration**:
- OS Ubuntu 18.04
- Python version 3.7
- ART version 1.1.0 on dev 1.2.0 branch
- TensorFlowV2 2.1.0
- Keras 2.2.3 with Tensorflow 1.13.1

# Checklist

# Note
New PR for #282

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   - pip install -q -r test_requirements.txt
   - pip install -q pylint pycodestyle
   - pip install -q torch==1.3.1+cpu torchvision==0.4.2+cpu -f https://download.pytorch.org/whl/torch_stable.html
+  - pip install cma==2.7.0
   - pip list
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,6 @@ install:
   - pip install -q -r test_requirements.txt
   - pip install -q pylint pycodestyle
   - pip install -q torch==1.3.1+cpu torchvision==0.4.2+cpu -f https://download.pytorch.org/whl/torch_stable.html
-  - pip install cma==2.7.0
   - pip list
 
 script:
--- a/art/attacks/__init__.py
+++ b/art/attacks/__init__.py
@@ -21,6 +21,10 @@ from art.attacks.evasion.universal_perturbation import UniversalPerturbation
 from art.attacks.evasion.virtual_adversarial import VirtualAdversarialMethod
 from art.attacks.evasion.zoo import ZooAttack
 
+# Added New Attacks
+from art.attacks.evasion.pixel_threshold import PixelAttack
+from art.attacks.evasion.pixel_threshold import ThresholdAttack
+
 from art.attacks.poisoning.poisoning_attack_svm import PoisoningAttackSVM
 
 from art.attacks.extraction.functionally_equivalent_extraction import FunctionallyEquivalentExtraction
--- a/art/attacks/__init__.py
+++ b/art/attacks/__init__.py
@@ -20,8 +20,6 @@ from art.attacks.evasion.spatial_transformation import SpatialTransformation
 from art.attacks.evasion.universal_perturbation import UniversalPerturbation
 from art.attacks.evasion.virtual_adversarial import VirtualAdversarialMethod
 from art.attacks.evasion.zoo import ZooAttack
-
-# Added New Attacks
 from art.attacks.evasion.pixel_threshold import PixelAttack
 from art.attacks.evasion.pixel_threshold import ThresholdAttack
 
--- a/art/attacks/evasion/__init__.py
+++ b/art/attacks/evasion/__init__.py
@@ -18,7 +18,5 @@ from art.attacks.evasion.spatial_transformation import SpatialTransformation
 from art.attacks.evasion.universal_perturbation import UniversalPerturbation
 from art.attacks.evasion.virtual_adversarial import VirtualAdversarialMethod
 from art.attacks.evasion.zoo import ZooAttack
-
-# Added New Attacks
 from art.attacks.evasion.pixel_threshold import PixelAttack
 from art.attacks.evasion.pixel_threshold import ThresholdAttack
--- a/art/attacks/evasion/__init__.py
+++ b/art/attacks/evasion/__init__.py
@@ -18,3 +18,7 @@ from art.attacks.evasion.spatial_transformation import SpatialTransformation
 from art.attacks.evasion.universal_perturbation import UniversalPerturbation
 from art.attacks.evasion.virtual_adversarial import VirtualAdversarialMethod
 from art.attacks.evasion.zoo import ZooAttack
+
+# Added New Attacks
+from art.attacks.evasion.pixel_threshold import PixelAttack
+from art.attacks.evasion.pixel_threshold import ThresholdAttack
--- a/art/attacks/evasion/pixel_threshold.py
+++ b/art/attacks/evasion/pixel_threshold.py
@@ -30,7 +30,6 @@ The Pixel Attack is a generalisation of One Pixel Attack.
     https://arxiv.org/abs/1906.06026
 """
 from __future__ import absolute_import, division, print_function, unicode_literals
-# import warnings
 import logging
 from itertools import product
 
@@ -62,26 +61,26 @@ class PixelThreshold(EvasionAttack):
     | Pixel and Threshold Attack Paper link:
         https://arxiv.org/abs/1906.06026
     """
-
     attack_params = EvasionAttack.attack_params + \
         ['th', 'es', 'targeted', 'verbose']
 
     def __init__(self, classifier, th, es, targeted, verbose):
         """
         Create a :class:`.PixelThreshold` instance.
-        :param classifier:    A trained classifier.
-        :type  classifier:    :class:`.Classifier`
-        :param th:            threshold value of the Pixel/ Threshold attack.
-                              A none indicates finding a minimum threshold.
-        :type  th:            `int` or `none`
-        :param es:            Indicates whether the attack uses CMAES (0) or
-                              DE (1) as Evolutionary Strategy
-        :type  es:            `int`
-        :param targeted:      Indicates whether the attack is targeted (True)
-                              or untargeted (False)
-        :type  targeted:      `bool`
-        :param verbose:       Indicates whether to print verbose messages of ES used
-        :type  verbose:       `bool`
+        
+        :param classifier: A trained classifier.
+        :type  classifier: :class:`.Classifier`
+        :param th: threshold value of the Pixel/ Threshold attack. th=None 
+                   indicates finding a minimum threshold.
+        :type  th: `int` or `none`
+        :param es: Indicates whether the attack uses CMAES (0) or  DE (1) as 
+                   Evolutionary Strategy
+        :type  es: `int`
+        :param targeted: Indicates whether the attack is targeted (True) or 
+                         untargeted (False)
+        :type  targeted: `bool`
+        :param verbose: Indicates whether to print verbose messages of ES used
+        :type  verbose: `bool`
         """
         super(PixelThreshold, self).__init__(classifier)
         self._project = True
@@ -158,7 +157,7 @@ class PixelThreshold(EvasionAttack):
                 while True:
                     image_result = []
                     threshold = (start + end) // 2
-                    success, trial_image_result = self.attack(
+                    success, trial_image_result = self._attack(
                         image, target_class, threshold)
                     if image_result or success:
                         image_result = trial_image_result
@@ -171,7 +170,7 @@ class PixelThreshold(EvasionAttack):
                     if end < start:
                         break
             else:
-                success, image_result = self.attack(
+                success, image_result = self._attack(
                     image, target_class, self.th)
             adv_x_best += [image_result]
 
@@ -187,57 +186,52 @@ class PixelThreshold(EvasionAttack):
         if y is not None:
             y = to_categorical(y, self.classifier.nb_classes())
 
-        rate_best = 100 * \
+        logger.info('Success rate of Attack: %.2f%%', 100 * \
             compute_success(self.classifier, x, y, adv_x_best,
-                            self.targeted, 1)
-        logger.info('Success rate of Attack: %.2f%%', rate_best)
+                            self.targeted, 1))
         return adv_x_best
 
     def predict(self, test_x):
         """
-        TODO: Write Comment
+        Wrapper for classifier.predict function.
         """
         if isinstance(self.classifier, PyTorchClassifier):
             test_x = np.swapaxes(test_x, 1, 3).astype(np.float32)
 
         return self.classifier.predict(test_x)
 
-    def predict_classes(self, adv_x, x, target_class):
+    def _attack_success(self, adv_x, x, target_class):
         """
-        TODO: Write Comment
-        """
-        predictions = self.predict(
-            self.perturb_image(adv_x, x))[:, target_class]
-        return predictions if not self.targeted else 1 - predictions
-
-    def attack_success(self, adv_x, x, target_class):
-        """
-        TODO: Write Comment
+        Checks whether the given pertubation `adv_x` for the image `img`
+        is successful. 
         """
         predicted_class = np.argmax(
             self.predict(
-                self.perturb_image(
+                self._perturb_image(
                     adv_x, x))[0])
         return bool((self.targeted and predicted_class == target_class) or (
             not self.targeted and predicted_class != target_class))
 
-    def attack(self, image, target_class, limit):
+    def _attack(self, image, target_class, limit):
         """
-        TODO: Write Comment
+        Attack the given image `image` with the threshold `limit` for the `target_class` 
+        which is true label for untargetted attack and targetted label for targetted attack. 
         """
-        bounds, initial = self.get_bounds(image, limit)
+        bounds, initial = self._get_bounds(image, limit)
 
         def predict_fn(x):
-            return self.predict_classes(x, image, target_class)
+            predictions = self.predict(
+                self._perturb_image(x, image))[:, target_class]
+            return predictions if not self.targeted else 1 - predictions
 
         def callback_fn(x, convergence=None):
 
             if self.es == 0:
-                if self.attack_success(x.result[0], image, target_class):
+                if self._attack_success(x.result[0], image, target_class):
                     raise Exception(
                         'Attack Completed :) Earlier than expected')
             else:
-                return self.attack_success(x, image, target_class)
+                return self._attack_success(x, image, target_class)
 
         if self.es == 0:
 
@@ -291,8 +285,8 @@ class PixelThreshold(EvasionAttack):
                 polish=False)
             adv_x = strategy.x
 
-        if self.attack_success(adv_x, image, target_class):
-            return True, self.perturb_image(adv_x, image)[0]
+        if self._attack_success(adv_x, image, target_class):
+            return True, self._perturb_image(adv_x, image)[0]
         else:
             return False, image
 
@@ -330,9 +324,9 @@ class PixelAttack(PixelThreshold):
                 verbose)
         self.type_attack = 0
 
-    def perturb_image(self, x, img):
+    def _perturb_image(self, x, img):
         """
-        TODO: Write Comment
+        Perturbs the given image `img` with the given perturbation `x`.
         """
         if x.ndim < 2:
             x = np.array([x])
@@ -344,9 +338,9 @@ class PixelAttack(PixelThreshold):
                 image[x_pos % image.shape[-3], y_pos % image.shape[-2]] = rgb
         return imgs
 
-    def get_bounds(self, img, limit):
+    def _get_bounds(self, img, limit):
         """
-        TODO: Write Comment
+        Define the bounds for the image `img` within the limits `limit`.
         """
         initial = []
         if self.es == 0:
@@ -392,7 +386,7 @@ class ThresholdAttack(PixelThreshold):
             targeted=False,
             verbose=False):
         """
-        Create a :class:`.PixelAttack` instance.
+        Create a :class:`.ThresholdAttack` instance.
         """
         super(
             ThresholdAttack,
@@ -404,9 +398,9 @@ class ThresholdAttack(PixelThreshold):
                 verbose)
         self.type_attack = 1
 
-    def perturb_image(self, x, img):
+    def _perturb_image(self, x, img):
         """
-        TODO: Write Comment
+        Perturbs the given image `img` with the given perturbation `x`.
         """
         if x.ndim < 2:
             x = np.array([x])
@@ -419,9 +413,9 @@ class ThresholdAttack(PixelThreshold):
                 image[i, j, k] = adv[count]
         return imgs
 
-    def get_bounds(self, img, limit):
+    def _get_bounds(self, img, limit):
         """
-        TODO: Write Comment
+        Define the bounds for the image `img` within the limits `limit`.
         """
 
         def bound_limit(value):
@@ -455,7 +449,6 @@ class ThresholdAttack(PixelThreshold):
 # TODO: Make the attack compatible with current version of SciPy Optimize
 # Differential Evolution
 
-
 """
 A slight modification to Scipy's implementation of differential evolution.
 To speed up predictions, the entire parameters array is passed to `self.func`,
@@ -470,6 +463,120 @@ https://github.com/scipy/scipy/blob/70e61dee181de23fdd8d893eaa9491100e2218d7/sci
 differential_evolution:The differential evolution global optimization algorithm
 Added by Andrew Nelson 2014
 """
+
+# Copyright (c) 2001, 2002 Enthought, Inc.
+# All rights reserved.
+# Copyright (c) 2003-2017 SciPy Developers.
+# All rights reserved.
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#   a. Redistributions of source code must retain the above copyright notice,
+#      this list of conditions and the following disclaimer.
+#   b. Redistributions in binary form must reproduce the above copyright
+#      notice, this list of conditions and the following disclaimer in the
+#      documentation and/or other materials provided with the distribution.
+#   c. Neither the name of Enthought nor the names of the SciPy Developers
+#      may be used to endorse or promote products derived from this software
+#      without specific prior written permission.
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS OR CONTRIBUTORS
+# BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+# OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+# SciPy bundles a number of libraries that are compatibly licensed.  We list
+# these here.
+# Name: Numpydoc
+# Files: doc/sphinxext/numpydoc/*
+# License: 2-clause BSD
+#   For details, see doc/sphinxext/LICENSE.txt
+# Name: scipy-sphinx-theme
+# Files: doc/scipy-sphinx-theme/*
+# License: 3-clause BSD, PSF and Apache 2.0
+#   For details, see doc/sphinxext/LICENSE.txt
+# Name: Six
+# Files: scipy/_lib/six.py
+# License: MIT
+#   For details, see the header inside scipy/_lib/six.py
+# Name: Decorator
+# Files: scipy/_lib/decorator.py
+# License: 2-clause BSD
+#   For details, see the header inside scipy/_lib/decorator.py
+# Name: ID
+# Files: scipy/linalg/src/id_dist/*
+# License: 3-clause BSD
+#   For details, see scipy/linalg/src/id_dist/doc/doc.tex
+# Name: L-BFGS-B
+# Files: scipy/optimize/lbfgsb/*
+# License: BSD license
+#   For details, see scipy/optimize/lbfgsb/README
+# Name: SuperLU
+# Files: scipy/sparse/linalg/dsolve/SuperLU/*
+# License: 3-clause BSD
+#   For details, see scipy/sparse/linalg/dsolve/SuperLU/License.txt
+# Name: ARPACK
+# Files: scipy/sparse/linalg/eigen/arpack/ARPACK/*
+# License: 3-clause BSD
+#   For details, see scipy/sparse/linalg/eigen/arpack/ARPACK/COPYING
+# Name: Qhull
+# Files: scipy/spatial/qhull/*
+# License: Qhull license (BSD-like)
+#   For details, see scipy/spatial/qhull/COPYING.txt
+# Name: Cephes
+# Files: scipy/special/cephes/*
+# License: 3-clause BSD
+#   Distributed under 3-clause BSD license with permission from the author,
+#   see https://lists.debian.org/debian-legal/2004/12/msg00295.html
+#   Cephes Math Library Release 2.8:  June, 2000
+#   Copyright 1984, 1995, 2000 by Stephen L. Moshier
+#   This software is derived from the Cephes Math Library and is
+#   incorporated herein by permission of the author.
+#   All rights reserved.
+#   Redistribution and use in source and binary forms, with or without
+#   modification, are permitted provided that the following conditions are met:
+#       * Redistributions of source code must retain the above copyright
+#         notice, this list of conditions and the following disclaimer.
+#       * Redistributions in binary form must reproduce the above copyright
+#         notice, this list of conditions and the following disclaimer in the
+#         documentation and/or other materials provided with the distribution.
+#       * Neither the name of the <organization> nor the
+#         names of its contributors may be used to endorse or promote products
+#         derived from this software without specific prior written permission.
+#   THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+#   ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+#   WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+#   DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+#   DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+#   (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+#   LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+#   ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#   (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+#   SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+# Name: Faddeeva
+# Files: scipy/special/Faddeeva.*
+# License: MIT
+#   Copyright (c) 2012 Massachusetts Institute of Technology
+#   Permission is hereby granted, free of charge, to any person obtaining
+#   a copy of this software and associated documentation files (the
+#   "Software"), to deal in the Software without restriction, including
+#   without limitation the rights to use, copy, modify, merge, publish,
+#   distribute, sublicense, and/or sell copies of the Software, and to
+#   permit persons to whom the Software is furnished to do so, subject to
+#   the following conditions:
+#   The above copyright notice and this permission notice shall be
+#   included in all copies or substantial portions of the Software.
+#   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+#   EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+#   MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+#   NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+#   LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+#   OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+#   WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 __all__ = ['differential_evolution']
 
--- a/art/attacks/evasion/pixel_threshold.py
+++ b/art/attacks/evasion/pixel_threshold.py
@@ -410,8 +410,8 @@ class ThresholdAttack(PixelThreshold):
         if x.ndim < 2:
             x = np.array([x])
         imgs = np.tile(img, [len(x)] + [1] * (x.ndim + 1))	
-        x = x.astype(int)	        x = x.astype(int)
-        for adv, image in zip(x, imgs):	        return x
+        x = x.astype(int)
+        for adv, image in zip(x, imgs):
             for count, (i, j, k) in enumerate(	
                     product(range(image.shape[-3]), range(image.shape[-2]),	
                             range(image.shape[-1]))):	
--- a/art/attacks/evasion/pixel_threshold.py
+++ b/art/attacks/evasion/pixel_threshold.py
@@ -44,9 +44,9 @@ from scipy.optimize import OptimizeResult, minimize
 from cma import CMAEvolutionStrategy, CMAOptions
 # from scipy.optimize import differential_evolution
 
-from art.classifiers.classifier import ClassifierGradients
+from art.classifiers import PyTorchClassifier
 from art.attacks.attack import EvasionAttack
-from art.utils import compute_success
+from art.utils import compute_success, to_categorical, check_and_transform_label_format
 
 logger = logging.getLogger(__name__)
 
@@ -84,19 +84,6 @@ class PixelThreshold(EvasionAttack):
         :type  verbose:       `bool`
         """
         super(PixelThreshold, self).__init__(classifier)
-        if not isinstance(classifier, ClassifierGradients):
-            raise (
-                TypeError(
-                    'For `' +
-                    self.__class__.__name__ +
-                    '` classifier must be an instance of '
-                    '`art.classifiers.classifier.ClassifierGradients`, '
-                    'the provided classifier is instance of ' +
-                    str(
-                        classifier.__class__.__bases__) +
-                    '. '
-                    ' The classifier needs to provide gradients.'))
-
         self._project = True
         kwargs = {'th': th, 'es': es, 'targeted': targeted, 'verbose': verbose}
         PixelThreshold.set_params(self, **kwargs)
@@ -140,6 +127,9 @@ class PixelThreshold(EvasionAttack):
         :return: An array holding the adversarial examples.
         :rtype: `np.ndarray`
         """
+        y = check_and_transform_label_format(
+            y, self.classifier.nb_classes(), return_one_hot=False)
+
         if y is None:
             if self.targeted:
                 raise ValueError(
@@ -147,14 +137,18 @@ class PixelThreshold(EvasionAttack):
                     'attack.'
                 )
             y = np.argmax(self.classifier.predict(x), axis=1)
+        else:
+            if len(y.shape) > 1:
+                y = np.argmax(y, axis=1)
 
         if self.th is None:
             logger.info('Performing minimal perturbation Attack.')
 
-        if np.amax(x) > 1:
-            self.bound_limit = 255
-        else:
-            self.bound_limit = 1
+        if isinstance(self.classifier, PyTorchClassifier):
+            x = np.swapaxes(x, 1, 3).astype(np.float32)
+
+        if np.max(x) <= 1:
+            x = x * 255.
 
         adv_x_best = []
         for image, target_class in zip(x, y):
@@ -177,21 +171,42 @@ class PixelThreshold(EvasionAttack):
                     if end < start:
                         break
             else:
-                success, image_result = self.attack(x, y, self.th)
+                success, image_result = self.attack(
+                    image, target_class, self.th)
             adv_x_best += [image_result]
 
         adv_x_best = np.array(adv_x_best)
+
+        if np.max(x) <= 1:
+            x = x / 255.
+
+        if isinstance(self.classifier, PyTorchClassifier):
+            x = np.swapaxes(x, 1, 3).astype(np.float32)
+            adv_x_best = np.swapaxes(adv_x_best, 1, 3).astype(np.float32)
+
+        if y is not None:
+            y = to_categorical(y, self.classifier.nb_classes())
+
         rate_best = 100 * \
             compute_success(self.classifier, x, y, adv_x_best,
                             self.targeted, 1)
         logger.info('Success rate of Attack: %.2f%%', rate_best)
         return adv_x_best
 
+    def predict(self, test_x):
+        """
+        TODO: Write Comment
+        """
+        if isinstance(self.classifier, PyTorchClassifier):
+            test_x = np.swapaxes(test_x, 1, 3).astype(np.float32)
+
+        return self.classifier.predict(test_x)
+
     def predict_classes(self, adv_x, x, target_class):
         """
         TODO: Write Comment
         """
-        predictions = self.classifier.predict(
+        predictions = self.predict(
             self.perturb_image(adv_x, x))[:, target_class]
         return predictions if not self.targeted else 1 - predictions
 
@@ -200,7 +215,7 @@ class PixelThreshold(EvasionAttack):
         TODO: Write Comment
         """
         predicted_class = np.argmax(
-            self.classifier.predict(
+            self.predict(
                 self.perturb_image(
                     adv_x, x))[0])
         return bool((self.targeted and predicted_class == target_class) or (
@@ -212,8 +227,8 @@ class PixelThreshold(EvasionAttack):
         """
         bounds, initial = self.get_bounds(image, limit)
 
-        def predict_fn(xs):
-            return self.predict_classes(xs, image, target_class)
+        def predict_fn(x):
+            return self.predict_classes(x, image, target_class)
 
         def callback_fn(x, convergence=None):
 
@@ -241,10 +256,10 @@ class PixelThreshold(EvasionAttack):
             else:
                 std = limit
 
-            es = CMAEvolutionStrategy(initial, std / 4, opts)
+            strategy = CMAEvolutionStrategy(initial, std / 4, opts)
 
             try:
-                es.optimize(
+                strategy.optimize(
                     predict_fn,
                     maxfun=max(
                         1,
@@ -253,16 +268,16 @@ class PixelThreshold(EvasionAttack):
                     len(bounds) *
                     100,
                     callback=callback_fn)
-            except Exception as e:
+            except Exception as exception:
                 if self.verbose:
-                    print(e)
+                    print(exception)
                 pass
 
-            adv_x = es.result[0]
+            adv_x = strategy.result[0]
 
         else:
 
-            es = differential_evolution(
+            strategy = differential_evolution(
                 predict_fn,
                 bounds,
                 disp=self.verbose,
@@ -274,7 +289,7 @@ class PixelThreshold(EvasionAttack):
                 atol=-1,
                 callback=callback_fn,
                 polish=False)
-            adv_x = es.x
+            adv_x = strategy.x
 
         if self.attack_success(adv_x, image, target_class):
             return True, self.perturb_image(adv_x, image)[0]
@@ -315,21 +330,21 @@ class PixelAttack(PixelThreshold):
                 verbose)
         self.type_attack = 0
 
-    def perturb_image(self, xs, img):
+    def perturb_image(self, x, img):
         """
         TODO: Write Comment
         """
-        if xs.ndim < 2:
-            xs = np.array([xs])
-        imgs = np.tile(img, [len(xs)] + [1] * (xs.ndim + 1))
-        xs = xs.astype(int)
-        for x, im in zip(xs, imgs):
-            for pixel in np.split(x, len(x) // (2 + im.shape[-1])):
+        if x.ndim < 2:
+            x = np.array([x])
+        imgs = np.tile(img, [len(x)] + [1] * (x.ndim + 1))
+        x = x.astype(int)
+        for adv, image in zip(x, imgs):
+            for pixel in np.split(adv, len(adv) // (2 + image.shape[-1])):
                 x_pos, y_pos, *rgb = pixel
-                im[x_pos % im.shape[-3], y_pos % im.shape[-2]] = rgb
+                image[x_pos % image.shape[-3], y_pos % image.shape[-2]] = rgb
         return imgs
 
-    def get_bounds(self, img, th):
+    def get_bounds(self, img, limit):
         """
         TODO: Write Comment
         """
@@ -340,24 +355,24 @@ class PixelAttack(PixelThreshold):
                 initial += [i, j]
                 for k in range(img.shape[-1]):
                     initial += [img[i, j, k]]
-                if count == th - 1:
+                if count == limit - 1:
                     break
                 else:
                     continue
             min_bounds = [0, 0]
             for _ in range(img.shape[-1]):
                 min_bounds += [0]
-            min_bounds = min_bounds * th
+            min_bounds = min_bounds * limit
             max_bounds = [img.shape[-3], img.shape[-2]]
             for _ in range(img.shape[-1]):
-                max_bounds += [self.bound_limit]
-            max_bounds = max_bounds * th
+                max_bounds += [255]
+            max_bounds = max_bounds * limit
             bounds = [min_bounds, max_bounds]
         else:
             bounds = [(0, img.shape[-3]), (0, img.shape[-2])]
             for _ in range(img.shape[-1]):
-                bounds += [(0, self.bound_limit)]
-            bounds = bounds * th
+                bounds += [(0, 255)]
+            bounds = bounds * limit
         return bounds, initial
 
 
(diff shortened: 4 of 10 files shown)


shashankkotyan:
Hi @beat-buesser, I passed the Travis CI test for the attacks and then I ran into errors in Travis CI while running my unit tests. 

The problem is, my attacks works for three-dimensional images having shape `(img_width, img_height, img_channel)` or if in batches 4-dimensional vectors having shape `(batch_size, img_width, img_height, img_channel). Therefore in case of MNIST, I would presume that the classifier works with the input of shape `(28,28,1)` rather than `(28,28)`. I checked the files used in the unit test and I saw the ART module uses MNIST with `(28,28)` shape (or that's what Travis CI job-log tells me)'.

Another problem is, I saw another test on IRIS dataset. This dataset doesn't come in purview of the attacks implemented in the white papers.

Though I have checked the attacks on CIFAR datasets on my local machine and they work as intended.

Please advise how to proceed with this issue.

beat-buesser:
@shashankkotyan That's great news!
I think all tests of ART using MNIST use the format (28, 28, 1) except for PyTorch which uses (1, 28, 28), but it could be that some tests have reasons to deviate from this. In any case, ART should work with any input shape as long as the provided model accepts it as input.
About the IRIS tests, only use them if it makes sense for your attack. If your attack only works for images then you don't need to test with IRIS. Use the existing tests as a guidance and select and adapt them to create optimal tests for your attacks.

shashankkotyan:
I see @beat-buesser. Thanks for mentioning that PyTorch uses the input shape (`1,28,28`) instead of (`28, 28, 1`). I now understand the error logs and have rectified the errors in the update.

shashankkotyan:
Hi @beat-buesser, I just saw the results from Travis CI and the results were a bit unexpected. 
If you see the logs for Job `#1758.1` from `line 1900` to `line 1933` it shows an error in testing knock off nets. 
While Pixel attack (`line 2974`) and Threshold attack (`line 3084` ) seems to be completed without failures. 
Also, [Codecov](https://codecov.io/gh/IBM/adversarial-robustness-toolbox/pull/287?src=pr&el=h1)  report seems strange to me because it shows the change in files to which I have not committed.

beat-buesser:
@shashankkotyan I have restarted the failed test run, I think it had failed because there is still some randomness in some of the existing tests.

shashankkotyan:
I see @beat-buesser, thanks for the explanation ! 
Now I am seeing successful pass in each of the 5 checks ! 

Let me know what further changes I have to do in the code, if any.

beat-buesser:
@shashankkotyan Great, thank you, we will review your pull request in the coming days.

.travis.yml
```diff
@@ -24,6 +24,7 @@ install:
   - pip install -q -r test_requirements.txt
   - pip install -q pylint pycodestyle
   - pip install -q torch==1.3.1+cpu torchvision==0.4.2+cpu -f https://download.pytorch.org/whl/torch_stable.html
+  - pip install cma==2.7.0
```
beat-buesser: Please move this dependency to test_requirements.txt.

art/attacks/__init__.py
```diff
@@ -21,6 +21,10 @@
 from art.attacks.evasion.virtual_adversarial import VirtualAdversarialMethod
 from art.attacks.evasion.zoo import ZooAttack
 
+# Added New Attacks
```
beat-buesser: Please remove this comment line and the empty line 23
```suggestion
```

art/attacks/evasion/__init__.py
```diff
@@ -18,3 +18,7 @@
 from art.attacks.evasion.universal_perturbation import UniversalPerturbation
 from art.attacks.evasion.virtual_adversarial import VirtualAdversarialMethod
 from art.attacks.evasion.zoo import ZooAttack
+
+# Added New Attacks
```
beat-buesser: Please remove these two lines.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+
+from art.attacks import PixelAttack
+from art.utils import get_labels_np_array
+
+logger = logging.getLogger(__name__)
+
+
+class TestPixelAttack(TestBase):
+    """
+    TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        super().setUpClass()
+
+        cls.n_test = 2
+        cls.x_test_mnist = cls.x_test_mnist[0:cls.n_test]
+        cls.y_test_mnist = cls.y_test_mnist[0:cls.n_test]
+
+    def test_keras_mnist(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        classifier = get_classifier_kr()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            False)
+
+    def test_tensorflow_mnist(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        classifier, sess = get_classifier_tf()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            False)
+
+    def test_pytorch_mnist(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        x_test = np.swapaxes(self.x_test_mnist, 1, 3).astype(np.float32)
```
beat-buesser:
```suggestion
        x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 28, 28, 1)).astype(np.float32)
```

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        classifier = get_classifier_pt()
+        self._test_attack(classifier, x_test, self.y_test_mnist, False)
+
+    def test_keras_mnist_targeted(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        classifier = get_classifier_kr()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            True)
+
+    def test_tensorflow_mnist_targeted(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        classifier, sess = get_classifier_tf()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            True)
+
+    def test_pytorch_mnist_targeted(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        """
+        x_test = np.swapaxes(self.x_test_mnist, 1, 3).astype(np.float32)
```
beat-buesser:
```suggestion
        x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 28, 28, 1)).astype(np.float32)

```

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        classifier = get_classifier_pt()
+        self._test_attack(classifier, x_test, self.y_test_mnist, True)
+
+    def _test_attack(self, classifier, x_test, y_test, targeted):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_pixel_attack.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            targets = y_test
+
+        for es in [0, 1]:
+
+            df = PixelAttack(classifier, th=64, es=es, targeted=targeted)
+            x_test_adv = df.generate(x_test_original, targets)
+
+            self.assertFalse((x_test == x_test_adv).all())
+            self.assertFalse((0.0 == x_test_adv).all())
```
beat-buesser: Please add a test for the expected actual values of `x_test_adv`. See example: https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/attacks/test_fast_gradient.py#L149

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+
+from art.attacks import ThresholdAttack
+from art.utils import get_labels_np_array
+
+logger = logging.getLogger(__name__)
+
+
+class TestThresholdAttack(TestBase):
+    """
+    TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        super().setUpClass()
+
+        cls.n_test = 2
+        cls.x_test_mnist = cls.x_test_mnist[0:cls.n_test]
+        cls.y_test_mnist = cls.y_test_mnist[0:cls.n_test]
+
+    def test_keras_mnist(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+
+class TestThresholdAttack(TestBase):
+    """
+    TODO: Write Comment
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        classifier = get_classifier_kr()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            False)
+
+    def test_tensorflow_mnist(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        classifier, sess = get_classifier_tf()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            False)
+
+    def test_pytorch_mnist(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        x_test = np.swapaxes(self.x_test_mnist, 1, 3).astype(np.float32)
+        classifier = get_classifier_pt()
+        self._test_attack(classifier, x_test, self.y_test_mnist, False)
+
+    def test_keras_mnist_targeted(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        classifier = get_classifier_kr()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            True)
+
+    def test_tensorflow_mnist_targeted(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        classifier, sess = get_classifier_tf()
+        self._test_attack(
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            True)
+
+    def test_pytorch_mnist_targeted(self):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+        """
+        x_test = np.swapaxes(self.x_test_mnist, 1, 3).astype(np.float32)
+        classifier = get_classifier_pt()
+        self._test_attack(classifier, x_test, self.y_test_mnist, True)
+
+    def _test_attack(self, classifier, x_test, y_test, targeted):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+            targets = y_test
+
+        for es in [0, 1]:
+
+            df = ThresholdAttack(classifier, th=64, es=es, targeted=targeted)
+            x_test_adv = df.generate(x_test_original, targets)
+
+            self.assertFalse((x_test == x_test_adv).all())
+            self.assertFalse((0.0 == x_test_adv).all())
+
```
beat-buesser: Please add two tests for the expected actual values of `x_test_adv` and `y_pred`. See example: https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/attacks/test_fast_gradient.py#L149

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            True)
+
+    def test_pytorch_mnist_targeted(self):
+        """
+        TODO: Write Comment
+        """
+        x_test = np.swapaxes(self.x_test_mnist, 1, 3).astype(np.float32)
```
beat-buesser:
```suggestion
        x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 28, 28, 1)).astype(np.float32)
```

tests/attacks/test_threshold_attack.py
```diff
@@ -0,0 +1,170 @@
...
+            classifier,
+            self.x_test_mnist,
+            self.y_test_mnist,
+            False)
+
+    def test_pytorch_mnist(self):
+        """
+        TODO: Write Comment
+        """
+        x_test = np.swapaxes(self.x_test_mnist, 1, 3).astype(np.float32)
```
beat-buesser:
```suggestion
        x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 28, 28, 1)).astype(np.float32)
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+The Pixel Attack is a generalisation of One Pixel Attack.
+
+| One Pixel Attack Paper link:
+    https://ieeexplore.ieee.org/abstract/document/8601309/citations#citations
+    (arXiv link: https://arxiv.org/pdf/1710.08864.pdf)
+| Pixel and Threshold Attack Paper link:
+    https://arxiv.org/abs/1906.06026
+"""
+from __future__ import absolute_import, division, print_function, unicode_literals
+# import warnings
```
beat-buesser:
```suggestion
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+
+import numpy as np
+
+from scipy._lib.six import xrange, string_types
+from scipy._lib._util import check_random_state
+from scipy.optimize.optimize import _status_message
+from scipy.optimize import OptimizeResult, minimize
+
+from cma import CMAEvolutionStrategy, CMAOptions
+# from scipy.optimize import differential_evolution
```
beat-buesser:
```suggestion
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+
+        rate_best = 100 * \
+            compute_success(self.classifier, x, y, adv_x_best,
+                            self.targeted, 1)
+        logger.info('Success rate of Attack: %.2f%%', rate_best)
+        return adv_x_best
+
+    def predict(self, test_x):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        """
+        if isinstance(self.classifier, PyTorchClassifier):
+            test_x = np.swapaxes(test_x, 1, 3).astype(np.float32)
+
+        return self.classifier.predict(test_x)
+
+    def predict_classes(self, adv_x, x, target_class):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        """
+        predictions = self.predict(
+            self.perturb_image(adv_x, x))[:, target_class]
+        return predictions if not self.targeted else 1 - predictions
+
+    def attack_success(self, adv_x, x, target_class):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        predicted_class = np.argmax(
+            self.predict(
+                self.perturb_image(
+                    adv_x, x))[0])
+        return bool((self.targeted and predicted_class == target_class) or (
+            not self.targeted and predicted_class != target_class))
+
+    def attack(self, image, target_class, limit):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+                classifier,
+                th,
+                es,
+                targeted,
+                verbose)
+        self.type_attack = 0
+
+    def perturb_image(self, x, img):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        x = x.astype(int)
+        for adv, image in zip(x, imgs):
+            for pixel in np.split(adv, len(adv) // (2 + image.shape[-1])):
+                x_pos, y_pos, *rgb = pixel
+                image[x_pos % image.shape[-3], y_pos % image.shape[-2]] = rgb
+        return imgs
+
+    def get_bounds(self, img, limit):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+                classifier,
+                th,
+                es,
+                targeted,
+                verbose)
+        self.type_attack = 1
+
+    def perturb_image(self, x, img):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        for adv, image in zip(x, imgs):
+            for count, (i, j, k) in enumerate(
+                    product(range(image.shape[-3]), range(image.shape[-2]),
+                            range(image.shape[-1]))):
+                image[i, j, k] = adv[count]
+        return imgs
+
+    def get_bounds(self, img, limit):
+        """
+        TODO: Write Comment
```
beat-buesser: Please update docstring.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            else:
+                bounds += [bound]
+        if self.es == 0:
+            bounds = [minbounds, maxbounds]
+
+        return bounds, initial
+
+
+# TODO: Make the attack compatible with current version of SciPy Optimize
+# Differential Evolution
```
beat-buesser: Is this comment required?

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            y = np.argmax(self.classifier.predict(x), axis=1)
+        else:
+            if len(y.shape) > 1:
+                y = np.argmax(y, axis=1)
+
+        if self.th is None:
+            logger.info('Performing minimal perturbation Attack.')
+
+        if isinstance(self.classifier, PyTorchClassifier):
+            x = np.swapaxes(x, 1, 3).astype(np.float32)
```
beat-buesser: Please remove these lines, because `x` is provided by the user in the correct format.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            adv_x_best += [image_result]
+
+        adv_x_best = np.array(adv_x_best)
+
+        if np.max(x) <= 1:
+            x = x / 255.
+
+        if isinstance(self.classifier, PyTorchClassifier):
+            x = np.swapaxes(x, 1, 3).astype(np.float32)
+            adv_x_best = np.swapaxes(adv_x_best, 1, 3).astype(np.float32)
```
beat-buesser: Please remove these lines, because `x` and `adv_x_best` is provided by the user in the correct format.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+
+        if y is not None:
+            y = to_categorical(y, self.classifier.nb_classes())
+
+        rate_best = 100 * \
```
beat-buesser: Please integrate the calculation of `rate_best` into `logger.info` in line 193, that way it is only calculated if the logging information is actually logged.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            compute_success(self.classifier, x, y, adv_x_best,
+                            self.targeted, 1)
+        logger.info('Success rate of Attack: %.2f%%', rate_best)
+        return adv_x_best
+
+    def predict(self, test_x):
```
beat-buesser: Please replace the use and definition of this function with `classifier.predict`.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            bounds = [minbounds, maxbounds]
+
+        return bounds, initial
+
+
+# TODO: Make the attack compatible with current version of SciPy Optimize
+# Differential Evolution
+
+
+"""
```
beat-buesser: Please add the license of the original code to this docstring: https://github.com/scipy/scipy/blob/70e61dee181de23fdd8d893eaa9491100e2218d7/LICENSE.txt

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+    def predict(self, test_x):
+        """
+        TODO: Write Comment
+        """
+        if isinstance(self.classifier, PyTorchClassifier):
+            test_x = np.swapaxes(test_x, 1, 3).astype(np.float32)
+
+        return self.classifier.predict(test_x)
+
+    def predict_classes(self, adv_x, x, target_class):
```
beat-buesser:
```suggestion
    def _predict_classes(self, adv_x, x, target_class):
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        """
+        TODO: Write Comment
+        """
+        predictions = self.predict(
+            self.perturb_image(adv_x, x))[:, target_class]
```
beat-buesser:
```suggestion
            self._perturb_image(adv_x, x))[:, target_class]
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        return predictions if not self.targeted else 1 - predictions
+
+    def attack_success(self, adv_x, x, target_class):
```
beat-buesser:
```suggestion
    def _attack_success(self, adv_x, x, target_class):
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        TODO: Write Comment
+        """
+        predicted_class = np.argmax(
+            self.predict(
+                self.perturb_image(
+                    adv_x, x))[0])
+        return bool((self.targeted and predicted_class == target_class) or (
+            not self.targeted and predicted_class != target_class))
+
+    def attack(self, image, target_class, limit):
```
beat-buesser:
```suggestion
    def _attack(self, image, target_class, limit):
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            PixelAttack,
+            self).__init__(
+                classifier,
+                th,
+                es,
+                targeted,
+                verbose)
+        self.type_attack = 0
+
+    def perturb_image(self, x, img):
```
beat-buesser:
```suggestion
    def _perturb_image(self, x, img):
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            x = np.array([x])
+        imgs = np.tile(img, [len(x)] + [1] * (x.ndim + 1))
+        x = x.astype(int)
+        for adv, image in zip(x, imgs):
+            for pixel in np.split(adv, len(adv) // (2 + image.shape[-1])):
+                x_pos, y_pos, *rgb = pixel
+                image[x_pos % image.shape[-3], y_pos % image.shape[-2]] = rgb
+        return imgs
+
+    def get_bounds(self, img, limit):
```
beat-buesser:
```suggestion
    def _get_bounds(self, img, limit):
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+            ThresholdAttack,
+            self).__init__(
+                classifier,
+                th,
+                es,
+                targeted,
+                verbose)
+        self.type_attack = 1
+
+    def perturb_image(self, x, img):
```
beat-buesser:
```suggestion
    def _perturb_image(self, x, img):
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1346 @@
...
+        imgs = np.tile(img, [len(x)] + [1] * (x.ndim + 1))
+        x = x.astype(int)
+        for adv, image in zip(x, imgs):
+            for count, (i, j, k) in enumerate(
+                    product(range(image.shape[-3]), range(image.shape[-2]),
+                            range(image.shape[-1]))):
+                image[i, j, k] = adv[count]
+        return imgs
+
+    def get_bounds(self, img, limit):
```
beat-buesser:
```suggestion
    def _get_bounds(self, img, limit):
```

tests/attacks/test_pixel_attack.py
shashankkotyan: Actually, as the prediction is calculated by the Evolutionary Optimisation which is stochastic in nature. The expected prediction will differ from machine to machine. Therefore, I did not add the expected predicted soft-labels (logits).

tests/attacks/test_threshold_attack.py
shashankkotyan: Actually, as the prediction is calculated by the Evolutionary Optimisation which is stochastic in nature. The expected prediction will differ from machine to machine. Therefore, I did not add the expected predicted soft-labels (logits).

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Kindly let it be here. The current scipy's version of differential evolution doesn't work but maybe in future I or someone else can remove the code included below the method and use this line. 
This line will act as a reminder to update the code.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Yes, The current version of Scipy's differential evolution doesn't work with the code and the following code which is a modification of Scipy's code works. However, it may be possible someone identifies the difference in the implementation and can resolve the usage by simply using the scipy's import.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: No, actually, when PyTorch is used the shape is `(img_channel, img_rows, img_cols) or channels_first`, however, my code requires the image in `channels_last` format. Therefore, I check if the PyTorch is used.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: No, actually, when PyTorch is used the shape is `(img_channel, img_rows, img_cols)`, however, my code requires the image in channels_last format. Therefore, I check if the PyTorch is used. 
Please note that I change x to `channels_last` format and similarly the adv_x generated is in `channels_last` format. Therefore I need to change it again to `channels_first` format for the PyTorch.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: No, actually, when PyTorch is used the shape is (img_channel, img_rows, img_cols), however, my code requires the image in channels_last format. Therefore, I check if the PyTorch is used.
Please note that I change x to channels_last format and similarly the adv_x generated is in channels_last format. Therefore I need to change it again to channels_first format for the PyTorch.
Therefore to use `classifier.predict`, I need to change it according to the format required by the classifier. The function itself corrected the image format for the classifier.predict function.

.travis.yml
shashankkotyan: Moved.

art/attacks/__init__.py
shashankkotyan: Removed.

art/attacks/evasion/__init__.py
shashankkotyan: Removed.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: I could not find a sample docstring for the given function.

tests/attacks/test_threshold_attack.py
shashankkotyan: I could not find a sample docstring for the given function.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Actually the code uses a modified script of SciPy's implementation of Differential Evolution Optimisation. However, the best case would be to make the DE optimisation compatible with the current SciPy's implementation.
At the current stage, I have not yet identified all the changes in the implementation but it can be done at a later stage. 

_Please let it be here so that later I or someone else can update this._

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Actually the code uses a modified script of SciPy's implementation of Differential Evolution Optimisation. However, the best case would be to make the DE optimisation compatible with the current SciPy's implementation.
At the current stage, I have not yet identified all the changes in the implementation but it can be done at a later stage.

_Please let it be here so that later I or someone else can update this. This comment ideally would hint that a version of SciPy's Differential Evolution is used._

.travis.yml
shashankkotyan: Moved the dependency.

art/attacks/__init__.py
shashankkotyan: Removed Unwanted lines.

art/attacks/evasion/__init__.py
shashankkotyan: Removed Unwanted Lines

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Removed.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Removed the function.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Added.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
shashankkotyan: Please note that the Pixel Attack is a black-box attack which uses Differential Evolution or Covariance Matrix Adaptation Evolution Strategy optimisations. Both of these optimisations have a bit of stochastics, which is why the perturbation can be different due to which the expected soft-label vector which is hard to predict.
In my experiments we have found there exists multiple solutions (perturbations) and we do not differentiate between them. However, if there was only a single solution possible, in that case, we would have a definite `y_test_pred_expected` value for that solution. 

In a gist, there exist multiple solutions and we don't differentiate between them so there is no expectation for the predicted y_test.

tests/attacks/test_threshold_attack.py
shashankkotyan: Please note that the Threshold Attack is a black-box attack which uses Differential Evolution or Covariance Matrix Adaptation Evolution Strategy optimisations. Both of these optimisations have a bit of stochastics, which is why the perturbation can be different due to which the expected soft-label vector which is hard to predict.
In my experiments we have found there exists multiple solutions (perturbations) and we do not differentiate between them. However, if there was only a single solution possible, in that case, we would have a definite `y_test_pred_expected` value for that solution. 

In a gist, there exist multiple solutions and we don't differentiate between them so there is no expectation for the predicted y_test.

tests/attacks/test_pixel_attack.py
shashankkotyan: Please note that the `get_classifier_pt` expects an input of shape `(1,28,28)` and not `(28,28,1)`
The suggestion would, therefore, create a contradiction to the expected shape.

https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/utils_test.py#L570

tests/attacks/test_pixel_attack.py
shashankkotyan: Please note that the `get_classifier_pt` expects an input of shape `(1,28,28)` and not `(28,28,1)`
The suggestion would, therefore, create a contradiction to the expected shape.

https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/utils_test.py#L570

tests/attacks/test_threshold_attack.py
shashankkotyan: Please note that the `get_classifier_pt` expects an input of shape `(1,28,28)` and not `(28,28,1)`
The suggestion would, therefore, create a contradiction to the expected shape.

https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/utils_test.py#L570

tests/attacks/test_threshold_attack.py
shashankkotyan: Please note that the `get_classifier_pt` expects an input of shape `(1,28,28)` and not `(28,28,1)`
The suggestion would, therefore, create a contradiction to the expected shape.

https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/utils_test.py#L570

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Please note that the `get_classifier_pt` expects an input of shape `(1,28,28)` and not `(28,28,1)`
The suggestion would, therefore, create a contradiction to the expected shape.

https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/tests/utils_test.py#L570

shashankkotyan: Please note that the `PyTorchClassifier` expects an input of shape `(1,28,28) or Channels First` and not `(28,28,1) or Channels Last`. Therefore, input would be in the shape `(1,28,28) or Channels First`. However, my attacks use the format `(28,28,1) or Channels Last` and these lines change it to the required format.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Please note that the `PyTorchClassifier` expects an input of shape `(1,28,28) or Channels First` and not `(28,28,1) or Channels Last`. Therefore, input would be in the shape `(1,28,28) or Channels First`. However, my attacks use the format `(28,28,1) or Channels Last` and these lines change it to the required format. 
This function changes the format before the `classifier.predict` is called.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Please note that the attacks give an output of shape `(28,28,1) or Channels Last` and not `(1,28,28) or Channels First`. Therefore, Output would be in the shape `(28,28,1) or Channels Last`. However, `PyTorchClassifier` use the format `(1,28,28) or Channels First` and these lines change it to the required format.

shashankkotyan:
Hi, @beat-buesser I have resolved the changes suggested. 
The changes which are not yet resolved, I have commented the reason for not updating it in this iteration. 
Kindly see the comments attached to the reviews. 

About the full-scale test. Yes, I have tested it with TensorflowV2 and Keras classifiers and results are same as expected.

tests/attacks/test_threshold_attack.py
shashankkotyan: Updated.

tests/attacks/test_pixel_attack.py
beat-buesser: Sorry, could you please replace it with this line, this time the shape should be correct:
```
x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 1, 28, 28)).astype(np.float32)
```
We use reshape instead of swapping axes to make sure that the PyTorch testing classifier is predicting the same as the Keras and TensorFlow testing classifiers.

tests/attacks/test_pixel_attack.py
beat-buesser: Sorry, could you please replace it with this line, this time the shape should be correct:
```
x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 1, 28, 28)).astype(np.float32)
```
We use reshape instead of swapping axes to make sure that the PyTorch testing classifier is predicting the same as the Keras and TensorFlow testing classifiers.

tests/attacks/test_pixel_attack.py
beat-buesser: Ok, I see. The reason why I'm asking is that the current tests only check if the attack runs without error and that the adversarial examples are different from the benign samples. Would there be a way to test more accurately if the attack works as expected? for example the number of pixels that get changed, the maximum difference between benign and adversarial examples, etc.?

tests/attacks/test_threshold_attack.py
beat-buesser: Ok, I see. The reason why I'm asking is that the current tests only check if the attack runs without error and that the adversarial examples are different from the benign samples. Would there be a way to test more accurately if the attack works as expected? for example the number of pixels that get changed, the maximum difference between benign and adversarial examples, etc.?

tests/attacks/test_threshold_attack.py
beat-buesser: Sorry, my suggestion contains the wrong format, could you please replace it with this line:
```
x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 1, 28, 28)).astype(np.float32)
```

tests/attacks/test_threshold_attack.py
beat-buesser: Sorry, my suggestion contains the wrong format, could you please replace it with this line:
```
x_test = np.reshape(self.x_test_mnist, (self.x_test_mnist.shape[0], 1, 28, 28)).astype(np.float32)
```

art/attacks/evasion/pixel_threshold.py
beat-buesser: Ok, could you please add this explanation as comment after line 45?

art/attacks/evasion/pixel_threshold.py
beat-buesser: Can we combine explanation with the comment above to be add in line 45?

art/attacks/evasion/pixel_threshold.py
beat-buesser: Ok, in that case please check for the value of `classifier.channel_index` to be 1 or 3 instead of checking for `isinstance(self.classifier, PyTorchClassifier)` because your attack could also be used, for example, with `KerasClassifier` and channel_index=1.

art/attacks/evasion/pixel_threshold.py
beat-buesser: Ok, in that case please check for the value of `classifier.channel_index` to be 1 or 3 instead of checking for `isinstance(self.classifier, PyTorchClassifier)` because your attack could also be used, for example, with `KerasClassifier` and channel_index=1.

art/attacks/evasion/pixel_threshold.py
beat-buesser: Ok, in that case please check for the value of `classifier.channel_index` to be 1 or 3 instead of checking for `isinstance(self.classifier, PyTorchClassifier)` because your attack could also be used, for example, with `KerasClassifier` and channel_index=1.

beat-buesser:
@shashankkotyan Thank you very much for responding and acting on all review requests. Could you please take a look at my new comments above? Could you please point me at the part of you attack that requires `Channel Last`?

tests/attacks/test_pixel_attack.py
shashankkotyan: No Problem. I have updated the code with the same.

tests/attacks/test_pixel_attack.py
shashankkotyan: No Problem. I have updated the code with the same.

tests/attacks/test_threshold_attack.py
shashankkotyan: No Problem. I have updated the code with the same.

tests/attacks/test_threshold_attack.py
shashankkotyan: No Problem. I have updated the code with the same.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: I have added an explanation around the import. Please let me know if it works or needs to be modified.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Yeah Sure. That would be an ideal case I think.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: I see, indeed checking for the `classifier.channel_index` makes more sense then checking for `class: PyTorchClassifier`. 
I have updated the checking.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: I see, indeed checking for the `classifier.channel_index` makes more sense then checking for `class: PyTorchClassifier`. 
I have updated the checking.

The answer for the question is mentioned in the comment.

tests/attacks/test_threshold_attack.py
shashankkotyan: I see now what you expected. Please allow me to explain in detail about the examples and then about more accurate tests. 

In the case of the Pixel Attack, the number of pixels to be changed is defined by the variable `th` which is a required parameter for the attack other we find the minimum pixel by searching. Here, as we explicitly, define the `th` values; therefore, the attacks are coded such a way that search occurs in the given define space and not anywhere else. Again the difference between the benign samples and adversarial samples vary because each pixel can have maximum perturbation of 255 or minimum disruption of 0. For example, if I attack a sample with `th=10`. The image is perturbed such a way that only `10` pixels are modified but again getting the L<sub>1</sub> or  L<sub>2</sub> distance would be in a range and will vary for sample to sample. 

In the case of the Threshold Attack, all pixels can be changed to a defined upper limit `th` which is a required parameter for the attack other we find the minimum pixel by searching. Here, as we explicitly, define the `th` values; therefore, the attacks are coded such a way that search occurs in the given define space and not anywhere else. Again the difference between the benign samples and adversarial samples vary because maybe all pixels change or none changes. For example, if I attack a sample with `th=1`. The image is perturbed such a way that the pixels are modified in a way that all pixels have a maximum L<sub>1</sub> of `th` but again getting the L<sub>1</sub> or  L<sub>2</sub> distance for the entire image (adversarial sample) would be in a range and will vary for sample to sample. 

I don't think we need extensive tests to verify for the `th` values for the attacks because the evolutionary optimisation handles it and it defined explicitly in the function `get_bounds` for both the attacks. 

I would suggest adding the tests for the optimisation techniques if you require extensive tests.
Because if, the attack is not working as expected. It would be the optimisation which fails. However, I haven't encountered a test or a situation where optimisation didn't work as expected yet the attacks were working. I would run into errors if the optimisation didn't work.

tests/attacks/test_pixel_attack.py
shashankkotyan: I see now what you expected. Please allow me to explain in detail about the examples and then about more accurate tests. 

In the case of the Pixel Attack, the number of pixels to be changed is defined by the variable `th` which is a required parameter for the attack other we find the minimum pixel by searching. Here, as we explicitly, define the `th` values; therefore, the attacks are coded such a way that search occurs in the given define space and not anywhere else. Again the difference between the benign samples and adversarial samples vary because each pixel can have maximum perturbation of 255 or minimum disruption of 0. For example, if I attack a sample with `th=10`. The image is perturbed such a way that only `10` pixels are modified but again getting the L<sub>1</sub> or  L<sub>2</sub> distance would be in a range and will vary for sample to sample. 

In the case of the Threshold Attack, all pixels can be changed to a defined upper limit `th` which is a required parameter for the attack other we find the minimum pixel by searching. Here, as we explicitly, define the `th` values; therefore, the attacks are coded such a way that search occurs in the given define space and not anywhere else. Again the difference between the benign samples and adversarial samples vary because maybe all pixels change or none changes. For example, if I attack a sample with `th=1`. The image is perturbed such a way that the pixels are modified in a way that all pixels have a maximum L<sub>1</sub> of `th` but again getting the L<sub>1</sub> or  L<sub>2</sub> distance for the entire image (adversarial sample) would be in a range and will vary for sample to sample. 

I don't think we need extensive tests to verify for the `th` values for the attacks because the evolutionary optimisation handles it and it defined explicitly in the function `get_bounds` for both the attacks. 

I would suggest adding the tests for the optimisation techniques if you require extensive tests.
Because if, the attack is not working as expected. It would be the optimisation which fails. However, I haven't encountered a test or a situation where optimisation didn't work as expected yet the attacks were working. I would run into errors if the optimisation didn't work.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: I see, indeed checking for the `classifier.channel_index` makes more sense then checking for `class: PyTorchClassifier`. 
I have updated the checking.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Integrated.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1457 @@
...
+            for _ in range(img.shape[-1]):
+                max_bounds += [255]
+            max_bounds = max_bounds * limit
+            bounds = [min_bounds, max_bounds]
+        else:
+            bounds = [(0, img.shape[-3]), (0, img.shape[-2])]
+            for _ in range(img.shape[-1]):
+                bounds += [(0, 255)]
+            bounds = bounds * limit
+        return bounds, initial
```
shashankkotyan: As requested, this is the first instance where input image shape is used. 
This function defines the bound limit for the pixel attack. 
Please note that the Pixel Attack perturbation is a vector of 5 variables for 3 channel image and 3 variable for 1 channel image, 
`For 3 Channel Index: <Position of X>, <Position of Y>, <Value of Red Channel>, <Value of Blue Channel>, <Value of Green Channel>`
`For 1Channel Index: <Position of X>, <Position of Y>, <Value of Monochrome Channel>'.

I tried to use `classifer.input_shape` but this parameter is not mandatory for all classifiers, therefore, I did not use it.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1457 @@
...
+        """
+        if x.ndim < 2:
+            x = np.array([x])
+        imgs = np.tile(img, [len(x)] + [1] * (x.ndim + 1))
+        x = x.astype(int)
+        for adv, image in zip(x, imgs):
+            for pixel in np.split(adv, len(adv) // (2 + image.shape[-1])):
+                x_pos, y_pos, *rgb = pixel
+                image[x_pos % image.shape[-3], y_pos % image.shape[-2]] = rgb
+        return imgs
```
shashankkotyan: As requested, this is the first instance where input image shape is used. 
This function adds the perturbation to the image. 
Please note that the Pixel Attack perturbation is a vector of 5 variables for 3 channel image and 3 variable for 1 channel image, 
`For 3 Channel Index: <Position of X>, <Position of Y>, <Value of Red Channel>, <Value of Blue Channel>, <Value of Green Channel>`
`For 1Channel Index: <Position of X>, <Position of Y>, <Value of Monochrome Channel>'.

I tried to use `classifer.input_shape` but this parameter is not mandatory for all classifiers, therefore, I did not use it.

shashankkotyan:
Hi @beat-buesser. I have seen the updated comments and have updated the code with the suggestions mentioned. I have left a few reviews yet to be resolved. Please take a look at the non-resolved reviews. 

About the part which requires `channels_last`. It would be the code of Pixel Attack class and its functions which are `_get_bounds` and `_perturb_image`. I would suggest you take a look at these two functions. 

It would help, if I had `classifier.input_shape` but I saw this parameter is not mandatory for all classifiers and therefore I didn't use it.

tests/attacks/test_pixel_attack.py
beat-buesser: Ok. My recommendation would still be to add more assertions in the tests to make sure that the attack keeps working as intended in the future, but I don't want to insist on it.

tests/attacks/test_threshold_attack.py
beat-buesser: Ok. My recommendation would still be to add more assertions in the tests to make sure that the attack keeps working as intended in the future, but I don't want to insist on it.

art/attacks/evasion/pixel_threshold.py
beat-buesser: Looks good.

art/attacks/evasion/pixel_threshold.py
beat-buesser: Ok.

beat-buesser:
@shashankkotyan Thank you for your updates and keeping working on the integration!

My biggest concern about the swapaxes in the attack is that if a classifier with channel_index=1 is trained on images that have been changed with reshape instead of swapaxes or in other words the attacks currently make an assumption on how the images are transformed before feeding them to the model. Therefore, it would be great if we could find a way to avoid swapaxes in the attacks.

`classifier.input_shape` is available in every classifier in ART and you can use it safely. I think the @abstractmethod is missing in the API because of a bug in combination with @property in earlier Python versions.

art/attacks/evasion/pixel_threshold.py
beat-buesser: If you would use `classifer.input_shape`, could you avoid `swapaxes`?

art/attacks/evasion/pixel_threshold.py
beat-buesser: If you would use `classifer.input_shape`, could you avoid `swapaxes`?

tests/attacks/test_pixel_attack.py
shashankkotyan: I see, I will think of some more assertions and add it in future. 
Currently, I can't think of any.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Indeed, if I initially have `classifer.input_shape`, I could definitely avoid `swapaxes`.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Indeed, if I initially have `classifer.input_shape`, I could definitely avoid `swapaxes`.

shashankkotyan:
I understand your concern regarding `swapaxes`. I agree it's not the most efficient way to handle `channel_index=1`. 

As you suggested `classifier.input_shape` is available in every classifier. However, I couldn't see the parameter in the `KerasClassifier` or `TensorflowClassifier` given by the user. Can you confirm that that `classifier.input_shape` is available?

beat-buesser:
That's great news! Yes, I can confirm that `classifier.input_shape` is available. For `KerasClassifier` the property value is set here https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/art/classifiers/keras.py#L126

and this property is accessed in all classifiers with the method inherited from the parent class `Classifier` defined here https://github.com/IBM/adversarial-robustness-toolbox/blob/e6d0b53a16bca37712a78ad916018c67a5951abe/art/classifiers/classifier.py#L120

shashankkotyan:
Thanks for the confirmation!
I will indeed update the code accordingly.

shashankkotyan:
Hi @beat-buesser, I have resolved the issue with `swapaxes`. 
However, once again the Travis CI gives error unrelated to the code.  

Let me know if I need to do some more changes.

beat-buesser:
Hi @shashankkotyan That's great!
I have checked the unittests and they are failing everywhere because of the recent release of XGBoost 1.0.0. I'm fixing it with pull request #302, once it is merged to dev_1.2.0 you can fix it by updating your branch with dev_1.2.0.

beat-buesser:
@shashankkotyan The branch `dev_1.2.0` is now ready to update your branch to fix the unittests.

shashankkotyan:
@beat-buesser The code passed the Travis-CI test.

beat-buesser:
@shashankkotyan That's really great! Thank you also for your patience! It looks good to me.

@mathsinn Could you please take a second review to make sure everything is ok?

art/attacks/evasion/pixel_threshold.py
```diff
@@ -0,0 +1,1458 @@
...
+        #         self.population_energies[candidate] = energy
+
+        #         # if the trial candidate also has a lower energy than the
+        #         # best solution then replace that as well
+        #         if energy < self.population_energies[0]:
+        #             self.population_energies[0] = energy
+        #             self.population[0] = trial
+
+        ##############
+        ##############
```
mathsinn: Could those comments be removed?

shashankkotyan: Hi @mathsinn, I am not sure whether removing these lines can be removed. All the lines in the file from `line number 454` is modified from SciPy's implementation of Differential Evolution. 
As you can see in the original implementation, these lines are used; however, in the modification, these lines are commented out. 

https://github.com/scipy/scipy/blob/70e61dee181de23fdd8d893eaa9491100e2218d7/scipy/optimize/_differentialevolution.py#L695

Therefore this comment comes under the changes done to the original code. 

I would prefer if you can keep this commented section, so people have a clear indication of how and what changes are done to the original code.

beat-buesser:
@shashankkotyan Thank you very much for sharing your great code with us and for your patience during the review process!

PR #304, opened 2020-02-21
Boundary and HopSkipJump attacks for unsquared images

### LINKED ISSUE #288: Boundary attack doesn't support unsquared images
**Describe the bug**
The current implementation of Boundary attack doesn't seem to support unsquared images, this issue will investigate and provide a solution for it.

### LINKED ISSUE #297: HopSkipJump doesn't support unsquared images
HopSkipJump attack doesn't seem to support unsquared images, this issue will investigate and provide a solution for it.
This issue is similar to #288.

minhitbk:
# Description
Fixes #288 
Fixes #297 

## Type of change

- [x] Bug fix (non-breaking)

# Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:
- OS
- Python version

# Checklist


art/attacks/evasion/boundary.py
```diff
@@ -300,8 +300,7 @@ def _orthogonal_perturb(self, delta, current_sample, original_sample):
         :rtype: `np.ndarray`
         """
         # Generate perturbation randomly
-        # input_shape = current_sample.shape
-        perturb = np.random.randn(*self.classifier.input_shape).astype(ART_NUMPY_DTYPE)
+        perturb = np.random.randn(*current_sample.shape).astype(ART_NUMPY_DTYPE)
```
beat-buesser: Why did you change from `classifier.input_shape` to `current_sample_shape`? Shouldn't they be the same?

art/attacks/evasion/boundary.py
```diff
@@ -315,7 +314,8 @@ def _orthogonal_perturb(self, delta, current_sample, original_sample):
             direction = np.swapaxes(direction, 0, self.classifier.channel_index - 1)
             for i in range(direction.shape[0]):
                 direction[i] /= np.linalg.norm(direction[i])
-                perturb[i] -= np.dot(perturb[i], direction[i]) * direction[i]
+                perturb[i] -= np.dot(np.dot(perturb[i], direction[i].T), direction[i])
```
beat-buesser: Is it correct that the changed line does the same as before and that the only change is that is uses explicitly numpy functions?

art/attacks/evasion/boundary.py
minhitbk: Yes, they are basically the same. But as we talked before, when the Keras classifier is not fixed with a prediction that doesn't check shape, inputshape maybe different than sample shape. And then it caused an error. So now, that problem was fixed so then this change is not needed anymore. Do you want to go back with the previous one or still keep this as they are not different.

art/attacks/evasion/boundary.py
minhitbk: Yes, this is confirmed with the notebooks' result of boundary attack, it works well with both untargeted and targeted attack on unsquared images.

minhitbk:
@beat-buesser: problem is fixed in these 2 lines:
             perturb[i] -= np.dot(perturb[i], direction[i]) * direction[i]
             perturb[i] -= np.dot(np.dot(perturb[i], direction[i].T), direction[i])

The second line accepts unsquared matrixes while the first line raises an error.

PR #325, opened 2020-03-05
Update method fit_generator in all classifiers

### LINKED ISSUE #323: Check method fit_generator in KerasClassifier
The method fit_generator is currently commented out. It should be un-commented and tested.

beat-buesser:
# Description

This pull request updates the method `fit_generator` in all classifiers. It updates the conditions to use framework fit functions for generators and applies reduce_labels for loss functions requiring index labels.

Fixes #323 

## Type of change

- [x] Bug fix (non-breaking)


minhitbk:
Hi Beat, I have a number of minor concerns as follows:
1. Why do we need to implement a fit_generator function for each of the framework instead of calling directly to that function in the parent classifier?
2. Should we also include preprocessing and preprocessing defences into the DataGenerator?
3. Calling to check reduce_labels should take into account the fact that a DataGenerator could produce index label.
4. In the fit_generator function of TensorflowV2, we should not call generator.data, but should utilize the function of generator.get_batch.

beat-buesser:
Hi @minhitbk Thank you very much for your great review!
1. I think the main idea of the framework-specific implementations is to take full advantage of optimised generator fit functions provided by some frameworks in cases where no defences and preprocessing are applied, and fall back to a numpy-based implementation it the parent class to apply defences and preprocessing.
2. I think that's a very important idea. I think it might be possible, but it probably requires a bit more work on the defences side, therefore I would like to propose to go ahead with the current approach for 1.2 and schedule this item for the 1.3 release.
3. I agree. I will implement this in this pull request.
4. I agree. I will implement this in this pull request.

beat-buesser:
@seanpmorgan Thank you for your review!

PR #350, opened 2020-03-20
Implement Estimator API

### LINKED ISSUE #316: Implement a prototype of the new Estimator API
Starting with epic #308 this issue will create a first prototype for the new estimator API on branch `development_estimator`.

beat-buesser:
# Description

This pull request implements the new Estimator API for ART. The main goals are to provide a new model abstraction that can easily be extended to support any machine learning task like object detection, regression, etc. and includes the existing classifier API. As a starting example this pull request includes an implementation of a framework- and model-specific estimator for a Faster-RCNN object detection model in PyTorch and an example script demonstrating a simple FGSM attack on this object detector. This pull request adapts the evasion attacks FGSM, BIM and PGD for the new estimator API while the other attacks currently remain limited to classifiers. Future pull request will update these other attacks where possible and provide improved checks and warnings.

Fixes #316 

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


tests/attacks/conftest.py
```diff
@@ -49,9 +49,9 @@ def get_image_classifier_list_for_attack(attack, defended=False, **kwargs):
             return None
 
         return [
-            potential_classier
-            for potential_classier in classifier_list
-            if attack.is_valid_classifier_type(potential_classier)
+            potential_classifier
```
killianlevacher, line 52: potential_classifier should probably be renamed as potential_estimator I guess?

tests/attacks/conftest.py
```diff
@@ -65,9 +65,9 @@ def _tabular_classifier_list(attack, clipped=True):
...
+            for potential_classifier in classifier_list
+            if all(t in type(potential_classifier).__mro__ for t in attack.estimator_requirements)
```
killianlevacher, line 70: I find this line a bit hard to read. this is just a suggestion and I'm not sure maybe if it's because I'm not so used to the Python way of doing things but would it not be simpler in terms of readability to have this line as a function instead? as in `attack.passes_estimator_requirements(potential_classifier)`?

tests/poison_detection/test_activation_defence.py
```diff
@@ -158,8 +158,8 @@ def test_analyze_cluster(self):
 
         self.defence.analyze_clusters(cluster_analysis="silhouette-scores")
 
-        report, dist_clean_by_class = self.defence.analyze_clusters(cluster_analysis="distance")
-        n_classes = self.classifier.nb_classes()
+        report, dist_clean_by_class = self.defence.analyze_clusters(cluster_analysis='distance')
```
killianlevacher, line 161: is it better to use 'these quotes' rather than "these quotes"? just asking out of curiosity :-)

tests/attacks/conftest.py
beat-buesser: That's a good catch. I have left all testing functions with classifier instead of changing to estimator because for now we are still testing only with classifiers in the new system and I'm not sure yet for which functions we will need new functions for estimators here. We will update these names as soon as we introduce tests for attacks using estimators for testing.

tests/attacks/conftest.py
beat-buesser: I agree. This is the Black style formatting (https://black.readthedocs.io/en/stable/the_black_code_style.html) that we are applying since ART 1.2. Some formatting can take time to get used to, but it is really consistent across all files.

tests/poison_detection/test_activation_defence.py
beat-buesser: This is also Black style formatting. We can continue using ' if it is easier to use on the keyboard and Black style formatting will update it automatically at the latest before each release.

PR #377, opened 2020-04-16
Create Docker image for ART

### LINKED ISSUE #370: Streamlining ART setup process
The current ART setup and development environment could benefit some standardization and clarification for both new comers and existing contributors.  

* For instance, as it stands, many contributors are using different versions of the pip packages requirements  which can lead to confusion and time wasted when debugging potentially platform specific bugs. 
* Also some packages (such as xgboost) seem to be causing problems on some platforms (ie: MacOS) but not on others.
* Readme documentation could also be improved to document various conventions used within the code base (such as where the datasets are downloaded)  and conventions used within ART tests

1. A dockerised version of ART which would provide a standardized running environment across platforms when dealing platform specific bugs
2. Separating the dataset downloads from the tests runs and/or mentioning to new users that various datasets should be downloaded prior to running the tests AND potentially changing/removing the use of `Config.jsons`
3. removing duplicate packages from  `requirements.txt` and `test_requirements.txt` files
4. making sure `requirement` packages all have a specified version associated with them (to guarantee a specific combination of packages does pass tests on travis)
5. letting users install only their framework of choice (and not all of them) in order to use or test the ART framework

killianlevacher:
# Description

Tried to address as many setup issues as possible to make things clearer for new comers. Also Dockerised the ART project to make sure we have a consistent reproducible environment across platforms. Updated README documentation accordingly.

Fixes #370

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


.gitignore
```diff
@@ -1,3 +1,6 @@
+#Adversarial Robustness Toolbox files
+datasets/
```
beat-buesser, line 2: What is the purpose of the directory `datasets`?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN pip3 list --outdated
+
+EXPOSE 8888
+
+CMD bash run_tests.sh
+
+#Check the Dockerfile here https://www.fromlatest.io/#/
+
+#NOTE to contributors: When changing/adding packages, please make sure that the packages are consitent with those
+# present within the requirements.txt and test_requirements.txt files
```
beat-buesser, line 34: Please add a newline at the end of this file.

README.md
```diff
@@ -165,24 +181,30 @@ python examples/<example_name>.py
...
-
+adding `Signed-off-By: Name<Email>` in the commit message. Example:
 ```bash
-git commit -s -m 'Add new feature'
+git commit -s -m 'Informative commit message'
 ```
 
+#### Unit tests
+When submitting additional unit tests for ART, in order to keep the code base maintainable, please make sure each unit 
+test can run in less than 10 seconds.
```
beat-buesser, line 204: I'm not sure if we can limit everything to 10 seconds. How about if we say `ideally within a few seconds`?

download_datasets.py
```diff
@@ -0,0 +1,13 @@
+import logging
```
beat-buesser, line 1: What is the purpose of this file?

requirements.txt
```diff
@@ -1,6 +1,9 @@
-matplotlib
-numpy
-scipy
+matplotlib==3.2.1
+numpy==1.18.1
+scipy==1.4.1
```
beat-buesser, line 3: Do we really need to define the requirements that tight?

test_requirements.txt
```diff
@@ -1,22 +1,26 @@
...
 
-keras
-mxnet
-#torch>=1.2.0
-tensorflow
+#ART supported Frameworks
+tensorflow==2.1.0
+keras==2.3.1
+# tensorflow==1.15.2
+# keras==2.2.5
```
beat-buesser, line 12: This breaks the testing on Travis where we are running with different version combination of TensorFlow and Keras. Currently all test runs are running with the same combination of `tensorflow==2.1.0` and `keras==2.3.1`. Is this change required for the Docker image?

test_requirements.txt
```diff
@@ -1,22 +1,26 @@
...
 tensorflow_addons
-scikit-learn==0.22.1
+
+mxnet==1.6.0
 xgboost==1.0.0
 lightgbm==2.3.1
+#torch>=1.2.0
+torch==1.4.0
```
beat-buesser, line 19: I think we should remove or comment these lines. We are currently installing torch==1.3.1 in .travis.yml for testing on Travis because installing it here has led to errors. The current test runs of this PR have run with torch 1.3.1

.gitignore
killianlevacher: a sorry, yep that was an omission. It was just temporary while I was working on my side. I've removed this change thanks for pointing it out

Dockerfile
killianlevacher: done

README.md
killianlevacher: lol ye I wasn't sure either. I initially put it at 15 seconds but ye your suggestion is better it gets the point across without putting a hard metric

requirements.txt
killianlevacher: true, I've instead updated the requirements with >= instead of == so that we are not that tight but at the same time have the benefit of knowing what versions should work. I think it's really important that we do nevertheless specify at a minimum the combination of package versions that we know work because we all seem to be using different versions of each packages within the team.  So it makes it harder to know if the issue is coming from the version packages or something else

test_requirements.txt
killianlevacher: no that's fine, it shouldn't impact the docker image. I can keep tensorflow and keras without any version number if that helps fix travis, and instead just specify them within the dockerfile directly

killianlevacher:
Had to rebase the commit graph to fixed an issue with a sign off missing on a commit. But this fails to update the current PR commit graph so I'm closing this PR and reopening another one with all the commits signed off

beat-buesser:
Continuing on #386

PR #410, opened 2020-05-11
Implement Wasserstein distance metric

### LINKED ISSUE #407: Implement Wasserstein distance as a new metric
Implement the Wassterstein distance as a new metric in `art.metrics` following eq. 3 in the linked paper.

Paper link: https://arxiv.org/abs/1902.07906
GitHub link: https://github.com/locuslab/projected_sinkhorn

beat-buesser:
# Description

This pull request implements the Wasserstein distance as a metric based on `scipy.stats.wasserstein_distance`. This new metric automatically reshapes the input distributions to 1D-distributions and iterates over the batch dimension. 

Fixes #407

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


art/metrics/metrics.py
```diff
@@ -352,3 +352,44 @@ def clever_t(classifier, x, target_class, nb_batches, batch_size, radius, norm,
...
+
+    assert u_values.shape == v_values.shape
+    if u_weights is not None:
+        assert v_weights is not None
+    if u_weights is None:
+        assert v_weights is None
+    if u_weights is not None and v_weights is not None:
+        assert u_weights.shape == v_weights.shape
+    if u_weights is not None:
+        assert u_values.shape[0] == u_weights.shape[0]
```
hesseltuinhof, line 382: It seems to me that you can place the last two asserts in the first condition block and the second `if` can be replaced with an `else`.
```suggestion
    if u_weights is not None:
        assert v_weights is not None
        assert u_weights.shape == v_weights.shape
        assert u_values.shape[0] == u_weights.shape[0]
    else:
        assert v_weights is None
```

art/metrics/metrics.py
```diff
@@ -352,3 +352,44 @@ def clever_t(classifier, x, target_class, nb_batches, batch_size, radius, norm,
...
+
+    u_values = u_values.flatten().reshape(u_values.shape[0], -1)
+    v_values = v_values.flatten().reshape(v_values.shape[0], -1)
+
+    wd = np.zeros(u_values.shape[0])
+
+    for i in range(u_values.shape[0]):
+        if u_weights is None and v_weights is None:
+            wd[i] = wasserstein_distance(u_values[i], v_values[i])
+        elif u_weights is not None and v_weights is not None:
```
hesseltuinhof, line 392: Both conditions must be `True` following the assertions and if the `if` condition is `False`. Thus no need to check anymore, right?
```suggestion
        else:
```

art/metrics/metrics.py
```diff
@@ -352,3 +352,44 @@ def clever_t(classifier, x, target_class, nb_batches, batch_size, radius, norm,
...
+    if u_weights is not None:
+        assert u_values.shape[0] == u_weights.shape[0]
+
+    u_values = u_values.flatten().reshape(u_values.shape[0], -1)
+    v_values = v_values.flatten().reshape(v_values.shape[0], -1)
+
+    wd = np.zeros(u_values.shape[0])
+
+    for i in range(u_values.shape[0]):
+        if u_weights is None and v_weights is None:
```
hesseltuinhof, line 390: The second condition has already been checked above. Thus most likely not necessary anymore, right?
```suggestion
        if u_weights is None:
```

art/metrics/metrics.py
beat-buesser: Nice!

art/metrics/metrics.py
beat-buesser: I thought that by having the condition explicit it would be easier to read, otherwise one would have to always remember the asserts, but I also would like to have the asserts to prevent unexpected errors.

art/metrics/metrics.py
beat-buesser: You are right, but I thought that by having the condition explicit it would be easier to read, otherwise one would have to always remember the asserts, but I also would like to have the asserts to prevent unexpected errors.

Issue #408, opened 2020-05-07
Video preprocessor defenses


hesseltuinhof:
Similar to the audio data case, it would be great to have a basic set of preprocessor defenses available for video data. 

For example, one could adapt the current `JpegCompression` and `SpatialSmoothing` preprocessors such that they are able to handle video data on a frame-by-frame basis. Furthermore, we have recently introduced the FFmpeg dependency, which is a powerful tool that can also convert or compress video data. Building on top of FFmpeg, we could thus add a video compression preprocessor that compresses original video input by choosing a lower bit rate.

hesseltuinhof:
* [x] `JpegCompression` now supports video data

PR #415, opened 2020-05-15
Add support for video data to the spatial smoothing defense

hesseltuinhof:
# Description

Add the possibility to use the spatial smoothing preprocessor with video data. 

Related to #408

## Type of change

# Testing

Several tests were provided to handle the video data case. Furthermore, the current tests have been moved to the `pytest` framework.

**Test Configuration**:
- Fedora 31
- Python 3.6.10
- ART 1.3.0 development

# Checklist


art/defences/preprocessor/spatial_smoothing.py
```diff
@@ -92,22 +91,33 @@ def __call__(self, x, y=None):
...
-        size = [1] + [self.window_size] * (len(x.shape) - 1)
-        size[self.channel_index] = 1
-        size = tuple(size)
+        filter_size = [self.window_size] * x_ndim
+        # set filter_size at batch and channels indices to 1
+        filter_size[0] = 1
+        filter_size[self.channel_index] = 1
+        # set filter_size at temporal index to 1
+        if x_ndim == 5:
+            temporal_index = 2 if self.channel_index == 1 else 1
```
beat-buesser, line 110: What do think about adding a similar comment here (e.g. # check if NCFHW or NFHWC) and in the docstring for `param x` as in the JpegCompression defence (#412) to show to which formats the following lines belong?

art/defences/preprocessor/spatial_smoothing.py
```diff
@@ -92,22 +91,33 @@ def __call__(self, x, y=None):
...
+        # Note median_filter:
+        # * center pixel located lower right
+        # * if window size even, use larger value (e.g. median(4,5)=5)
+        result = median_filter(x, size=tuple(filter_size), mode="reflect")
 
-        result = ndimage.filters.median_filter(x, size=size, mode="reflect")
         if self.clip_values is not None:
             np.clip(result, self.clip_values[0], self.clip_values[1], out=result)
 
-        return result.astype(ART_NUMPY_DTYPE), y
```
beat-buesser, line 110: Did you check if after removing this case to `ART_NUMPY_DTYPE` this call still returns an array of format `ART_NUMPY_DTYPE` for all possible input parameter formats?

tests/defences/preprocessor/test_spatial_smoothing.py
```diff
@@ -0,0 +1,115 @@
...
+import numpy as np
+import pytest
+from numpy.testing import assert_array_equal
+
+from art.defences.preprocessor import SpatialSmoothing
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
```
beat-buesser, line 31: These test are testing if arrays of the same number are still of the same number after smoothing which would also be true if no smoothing is applied at all, which is ok for testing some aspects. Would it be possible to add a test for one parameter condition with an array that contains varying numbers to test if smoothing is happening?

art/defences/preprocessor/spatial_smoothing.py
hesseltuinhof: `scipy.ndimage.median_filter` will by default return an array of the same dtype as input.

tests/defences/preprocessor/test_spatial_smoothing.py
hesseltuinhof: Agreed! Returning by accident `x` instead of `result` would not fail the tests at the moment. Great catch :+1:

hesseltuinhof:
I have added a additional test case and added your suggested comment. Thank you @beat-buesser.

PR has been rebased :+1:

beat-buesser:
@hesseltuinhof Great work, thank you very much!

PR #467, opened 2020-06-12
fix thermometer encoding

### LINKED ISSUE #450: Thermometer encoding usage
Hi, I'm attempting to use the thermometer encoding preprocessing defence, but I'm confused about the usage procedure of it and I didn't see an example/notebook with that particular defence.

Here's a minimal example:
```python
import numpy as np
from PIL import Image
from tensorflow.keras.applications.xception import Xception
from art.classifiers import TensorFlowV2Classifier
from art.defences.preprocessor import ThermometerEncoding

target_image = Image.open('dragonfly.jpg')
target_image = target_image.resize((299, 299), resample=Image.LANCZOS)
target_image = np.array(target_image, dtype=np.float64)

model = Xception(weights='imagenet')
target_image = target_image / 255
defence = ThermometerEncoding(clip_values=(0, 1))
art_model = TensorFlowV2Classifier(model=model, nb_classes=1000, input_shape=(299, 299, 3), clip_values=(0, 1),
                                   preprocessing=(0.5, 0.5), preprocessing_defences=defence)
art_model.predict(np.array([target_image]))
```

Any image will do in place of `dragonfly.jpg`, so I'm not attaching it here.
I would expect to still be able to use the wrapped model normally, i.e. like this `art_model.predict(np.array([target_image]))` but instead I get the following traceback:

```
WARNING:tensorflow:Model was constructed with shape (None, 299, 299, 3) for input Tensor("input_1:0", shape=(None, 299, 299, 3), dtype=float32), but it was called on an input with incompatible shape (1, 299, 299, 30).
Traceback (most recent call last):
  File "...\Python\Python38\lib\contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "...\Python\Python38\lib\site-packages\tensorflow\python\keras\engine\base_layer_utils.py", line 443, in enter
    yield
  File "...\Python\Python38\lib\site-packages\tensorflow\python\keras\engine\base_layer.py", line 968, in __call__
    outputs = self.call(cast_inputs, *args, **kwargs)
  File "...\Python\Python38\lib\site-packages\tensorflow\python\keras\engine\network.py", line 717, in call
    return self._run_internal_graph(
  File "...\Python\Python38\lib\site-packages\tensorflow\python\keras\engine\network.py", line 888, in _run_internal_graph
    output_tensors = layer(computed_tensors, **kwargs)
  File "...\Python\Python38\lib\site-packages\tensorflow\python\keras\engine\base_layer.py", line 968, in __call__
    outputs = self.call(cast_inputs, *args, **kwargs)
  File "...\Python\Python38\lib\site-packages\tensorflow\python\keras\layers\convolutional.py", line 195, in call
    self._convolution_op = nn_ops.Convolution(
  File "...\Python\Python38\lib\site-packages\tensorflow\python\ops\nn_ops.py", line 1060, in __init__
    raise ValueError(
ValueError: number of input channels does not match corresponding dimension of filter, 30 != 3
```

*   OS Windows 10
*   Python version 3.8.1
*   ART version dev_1.3.0
*   TensorFlow 2.2.0

Please note that I am using the dev version currently, which I switched to due to #436, but I have also tested it with 1.2.0 and get the same output.

mathsinn:
Signed-off-by: Mathieu Sinn <mathsinn@ie.ibm.com>

# Description

This PR fixes two bugs in the thermometer encoding preprocessing defense:
- Forward passes are correctly computed now (cf the changes to the unit tests);
- The `estimate_gradients` function now correctly backpropagates to the original input domain.

Fixes #450 

## Type of change

- [x] Bug fix (non-breaking)

# Testing

- [x] Unit tests
- [x] Tests in a notebook assessing the effectiveness of `estimate_gradients` for end-to-end adaptive attacks, using the example provided by @maliwin in #450.

**Test Configuration**:
- MacOS
- Python 3.6
- ART 1.2 (dev_1.3.0 branch)
- TensorFlow 1.15.0 

# Checklist


mathsinn:
@beat-buesser There are two things we still need or may want to think about:

1. The input shape checks in the `loss_gradient` and `class_gradient` functions of `KerasClassifier` will prevent users from computing gradients on Keras classifiers equipped with the thermometer preprocessing defense. To use the example from issue #450, the problem is that the inputs `x` have shape `(32, 32, 3)` while the Keras model takes input of shape `(32, 32, 3*5)`, and this is the shape that the check in the gradient functions is trying to enforce.

2. We may also want to think about the `clip_values` in `ThermometerEncoding`. Currently, they are enforced to be `(0, 1)` in `set_params` so there may not be that much value in exposing them to the user. Moreover, the encoded feature values are `0` and `1`, so they lie within that interval by design. What perhaps would be more valuable is to allow the clip values to specify the range of the original feature domain, e.g. `(-1, 1)`, and then apply the thermometer encoding over that domain (currently it's assumed the original feature domain is `(0, 1)`).

PR #538, opened 2020-08-11
Add new membership inference attacks

abigailgold:
# Description

Added two membership inference attacks: one rule-based and one learned black-box attack. 
Added two Jupyter notebooks explaining how to apply both membership and attribute inference attacks.
Added dependency on pandas for preprocessing of nursery dataset used in new notebooks.

## Type of change

- [ ] Bug fix (non-breaking)
- [x] This change requires a documentation update

# Testing

Added tests for new membership attacks in tests/attacks/inference/test_membership_inference.py

**Test Configuration**:
- OS: MacOS
- Python version: 3.7
- TensorFlow 2.1 / Keras 2.3.1 / PyTorch 0.1 

# Checklist


requirements.txt
```diff
@@ -11,6 +11,7 @@ pydub==0.24.1
 resampy==0.2.2
 ffmpeg-python==0.2.0
 cma==2.7.0
+pandas==0.24.2
```
beat-buesser, line 14: Is there a reason for using pandas 0.24 instead of pandas 1.x?

art/estimators/estimator.py
```diff
@@ -512,6 +512,20 @@ def set_learning_phase(self, train: bool) -> None:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def loss(self, x: np.ndarray, y: np.ndarray, **kwargs) -> np.ndarray:
+        """
+        Compute the loss of the neural network for samples `x`.
```
beat-buesser, line 518: Estimators don't have to be neural networks.
```suggestion
        Compute the loss of the model for samples `x`.
```

art/estimators/classification/tensorflow.py
```diff
@@ -907,6 +920,43 @@ def class_gradient(self, x: np.ndarray, label: Union[int, List[int], None] = Non
...
+                "defined."
+            )
+        prev_reduction = self._loss_object.reduction
+        self._loss_object.reduction = tf.keras.losses.Reduction.NONE
+
+        # Apply preprocessing
+        x_preprocessed, _ = self._apply_preprocessing(x, y, fit=False)
+
+        if tf.executing_eagerly():
+            # with tf.GradientTape() as tape:
```
beat-buesser, line 946:
```suggestion
```

art/estimators/classification/tensorflow.py
```diff
@@ -907,6 +920,43 @@ def class_gradient(self, x: np.ndarray, label: Union[int, List[int], None] = Non
...
+            x_preprocessed_tf = tf.convert_to_tensor(x_preprocessed)
+            # tape.watch(x_preprocessed_tf)
```
beat-buesser, line 948:
```suggestion
```

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+
+import logging
+from typing import Optional
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.optim as optim
+from torch.utils.data.dataset import Dataset
+from torch.utils.data import DataLoader
```
beat-buesser, line 34: To make ART framework independent, these imports have to be moved inside of the methods where they are required.

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+
+from art.attacks import InferenceAttack
+from art.estimators.estimator import BaseEstimator, NeuralNetworkMixin
+from art.estimators.classification.classifier import ClassifierMixin, Classifier
+from art.utils import check_and_transform_label_format
+from sklearn.ensemble import RandomForestClassifier, GradientBoostingClassifier
```
beat-buesser, line 40: For PEP8, please move the import of `sklearn` after the import of `numpy`.

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+use_cuda = torch.cuda.is_available()
+
+
+def to_cuda(x):
+    if use_cuda:
+        x = x.cuda()
+    return x
+
+
+class AttackDataset(Dataset):
```
beat-buesser, line 53: Could you please add a short docstring for this class?

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+        return len(self.x1)
+
+    def __getitem__(self, idx):
+        if idx >= len(self.x1):
+            raise IndexError("Invalid Index")
+
+        return self.x1[idx], self.x2[idx], self.y[idx]
+
+
+class MembershipInferenceAttackModel(nn.Module):
```
beat-buesser, line 73: Could you please add a short docstring for this class?

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+class MembershipInferenceBlackBoxRuleBased(InferenceAttack):
+    """
+        Implementation of a simple, rule-based black-box membership inference attack.
+
+        This implementation uses the simple rule: if the model's prediction for a sample is correct, then it is a
+        member. Otherwise, it is not a member.
+    """
+    _estimator_requirements = [BaseEstimator]
```
beat-buesser, line 121: Does `MembershipInferenceBlackBoxRuleBased` require classification models? If yes, we should add `ClassifierMixin` to `_estimator_requirements`.

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+
+
+class MembershipInferenceBlackBox(InferenceAttack):
+    """
+        Implementation of a learned black-box membership inference attack.
+
+        This implementation can use as input to the learning process probabilities/logits or losses,
+        depending on the type of model and provided configuration.
+    """
+    _estimator_requirements = [BaseEstimator]
```
beat-buesser, line 160: Does MembershipInferenceBlackBoxRuleBased require classification models? If yes, we should add ClassifierMixin to _estimator_requirements.
```suggestion
    _estimator_requirements = [BaseEstimator, ClassifierMixin]
```

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+        return [1 if p == y[index] else 0 for index, p in enumerate(predictions)]
+
+
+class MembershipInferenceBlackBox(InferenceAttack):
+    """
+        Implementation of a learned black-box membership inference attack.
+
+        This implementation can use as input to the learning process probabilities/logits or losses,
+        depending on the type of model and provided configuration.
+    """
```
beat-buesser, line 159:
```suggestion
    """
    
    attack_params = InferenceAttack.attack_params + [
        "input_type",
        "attack_model_type",
        "attack_model",
    ]
```

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+                    self.attack_model = MembershipInferenceAttackModel(classifier.nb_classes, 1)
+                self.epochs = 100
+                self.bs = 100
+                self.lr = 0.0001
+            elif attack_model_type == 'rf':
+                self.attack_model = RandomForestClassifier()
+            elif attack_model_type == 'gb':
+                self.attack_model = GradientBoostingClassifier()
+            else:
+                raise ValueError("Illegal value for parameter `attack_model_type`.")
```
beat-buesser, line 205: After setting all attributes we should run method `_check_params` which contains type and value checks for each attribute if possible. A method `_check_params` needs to be implemented by this class.
```suggestion
                raise ValueError("Illegal value for parameter `attack_model_type`.")
        self._check_params()
```

art/attacks/inference/membership_inference.py
```diff
@@ -0,0 +1,346 @@
...
+        Infer sensitive properties (attributes, membership training records) from the targeted estimator. This method
+        should be overridden by all concrete inference attack implementations.
+
+        :param x: An array with reference inputs to be used in the attack.
+        :param y: Labels for `x`. This parameter is only used by some of the attacks.
+        :return: An array holding the inferred properties.
+        """
+        raise NotImplementedError
+        # if NeuralNetworkMixin in type(self.estimator).__mro__:
+        #     activations = self.estimator.get_activations(x, self.estimator.layer_names[-1], batch_size=self.bs)
```
beat-buesser, line 346: How is this class used?

art/estimators/classification/keras.py
```diff
@@ -237,7 +242,7 @@ def _initialize_params(
             "__name__" in dir(loss_function)
             and loss_function.__name__
             in ["categorical_hinge", "categorical_crossentropy", "binary_crossentropy", "kullback_leibler_divergence",]
-        ) or (self.is_tensorflow and flag_is_instance):
```
beat-buesser, line 240: Is removing `self.is_tensorflow` required for this pull request? I'm still concerned that it was required here.

art/estimators/generation/tensorflow.py
```diff
@@ -100,8 +100,18 @@ def __init__(
             # TODO do the same thing for all not None variables
         self._sess = sess
 
-    def loss(self, z, image_adv):
-        pass
```
beat-buesser, line 104: @killianlevacher Did you need this method `loss(self, z, image_adv)` for something? Can we safely replace it with the new method below?

tests/attacks/inference/test_membership_inference.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+logger = logging.getLogger(__name__)
+attack_train_ratio = 0.5
+num_classes_iris = 3
+num_classes_mnist = 10
+
+
+@pytest.fixture()
+def fix_get_mnist_subset(get_mnist_dataset):
```
killianlevacher, line 47: this fixture needs to be removed and instead the project-wide `get_default_mnist_subset` should be used

tests/attacks/inference/test_membership_inference.py
```diff
@@ -0,0 +1,363 @@
...
+        logging.warning("Couldn't perform this test because no classifier is defined")
+        return
+
+    (x_train, y_train), (x_test, y_test) = get_iris_dataset
+
+    for classifier in classifier_list:
+        # print(type(classifier).__name__)
+        attack = MembershipInferenceBlackBoxRuleBased(classifier)
+        # infer attacked feature
+        inferred_train = attack.infer(x_train, y_train)
```
killianlevacher, line 90: lines from 90 to 97 are pratically identical across all tests. For maintenance purposes, it would be better to use a `_backend_test_something` method was used. An example of this pattern can be found at `tests.attacks.evasion.test_fast_gradient`

tests/attacks/inference/test_membership_inference.py
```diff
@@ -0,0 +1,363 @@
...
+    classifier_list = get_tabular_classifier_list(MembershipInferenceBlackBox)
+    if not classifier_list:
+        logging.warning("Couldn't perform this test because no classifier is defined")
+        return
+
+    (x_train, y_train), (x_test, y_test) = get_iris_dataset
+    attack_train_size = int(len(x_train) * attack_train_ratio)
+    attack_test_size = int(len(x_test) * attack_train_ratio)
+
+    model_types = ['nn', 'rf', 'gb']
```
killianlevacher, line 136: the `model_types` parameters should be set as a pytest marker parameter. See `@pytest.mark.parametrize("norm", [np.inf, 1, 2])` in test `tests.attacks.evasion.test_fast-gradient.test_norm_images`

tests/attacks/inference/test_membership_inference.py
```diff
@@ -0,0 +1,363 @@
...
```
killianlevacher, line 165: same thing as above

requirements.txt
abigailgold: That's just the version I already had installed. I can test with version 1.1 and see if it works as well.

art/attacks/inference/membership_inference.py
abigailgold: It is not at the moment. It's just a placeholder, I didn't have time to implement it for this release. I guess I should remove it then, right?

art/estimators/classification/keras.py
abigailgold: Yes, it is required for the tests to pass. Look at the following elif (line 250), it does a very similar check (checks if loss_function is an instance of a particular class) but does not also force self.is_tensorflow to be true. When using "regular" keras (i.e., import keras and not import tensorflow.keras) these loss classes still exist and when using them without this change you get to the raise ValueError("Loss function not recognised.").

tests/attacks/inference/test_membership_inference.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+@pytest.mark.only_with_platform("keras")
+@pytest.mark.skipif(keras.__version__.startswith('2.2'), reason="requires Keras 2.3.0 or higher")
+def test_black_box_keras_loss(get_iris_dataset):
+    (x_train, y_train), (x_test, y_test) = get_iris_dataset
+    attack_train_size = int(len(x_train) * attack_train_ratio)
+    attack_test_size = int(len(x_test) * attack_train_ratio)
+
+    model = keras.models.Sequential()
```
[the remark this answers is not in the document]
abigailgold: So there is actually nothing special about the model itself, but I needed to be able to test with both a string-defined loss (e.g., loss='categorical_crossentropy') and a class-based loss (e.g., loss=keras.losses.CategoricalCrossentropy()). The  get_tabular_classifier_list fixture only provides the first option. So how can I add an additional keras tabular classifier that uses a class-based loss object?

tests/attacks/inference/test_membership_inference.py
abigailgold: So are you saying I should put the _backend_test_something method in attacks/utils.py? Even if it's only used in this test file? Or should I add it within test_membership_inference.py?

tests/attacks/inference/test_membership_inference.py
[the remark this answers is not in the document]
killianlevacher: a ok that makes sense yes I understand the reason now. @beat-buesser I guess the only way we could make this test framework independent is if we enabled `get_tabular_classifier_list` to get a loss input? That might be something to look towards in the future. At a minimum, I would suggest then maybe to put a comment explaining the reason why these local models were needed (to help out future refactoring). Also if these these keras models will be left, there would be a need to add a `skip_by_platform` marker to the test to make sure it isn't ran by other frameworks

tests/attacks/inference/test_membership_inference.py
killianlevacher: yes I think leaving it in `test_membership_inference.py` is probably a better idea since it's only used here

art/attacks/inference/membership_inference.py
beat-buesser: Yes, removing was my first thought because it does not implement methods. But if you would like to claim the name space and have plans to implement it in the very near future you can leave it, in that case I would ask to add a comment describing the plans.

abigailgold:
Opened a new pull request from a clean branch (573)

PR #549, opened 2020-08-13
Update shadow_attack.py

mcguires5:
# Description

In the examples for generating attacks, all data is passed in at once, if this is the case then why is the data being resized to length of data * batch size? I cannot make this attack work, I have a feeling this is due to incorrect shaping of the data here. This may not be the optimal change, but I think something is wrong here.

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:
- OS
- Python version

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### DIFF
--- a/art/attacks/evasion/shadow_attack.py
+++ b/art/attacks/evasion/shadow_attack.py
@@ -131,9 +131,8 @@ class ShadowAttack(EvasionAttack):
             raise ValueError("Unrecognized input dimension. Shadow Attack can only be applied to image data.")
 
         x = x.astype(ART_NUMPY_DTYPE)
-        x_batch = np.repeat(x, repeats=self.batch_size, axis=0).astype(ART_NUMPY_DTYPE)
-        x_batch = x_batch + np.random.normal(scale=self.sigma, size=x_batch.shape).astype(ART_NUMPY_DTYPE)
-        y_batch = np.repeat(y, repeats=self.batch_size, axis=0)
+        x_batch = x + np.random.normal(scale=self.sigma, size=x.shape).astype(ART_NUMPY_DTYPE)
+        y_batch = y
 
         perturbation = (
             np.random.uniform(


beat-buesser:
Hi @mcguires5 Thank you very much for you this patch! We will take a look as soon as possible.

Could you please provide more information about model, dataset, etc. that you are using?

mcguires5:
I have tried this with 3 different dataset/model combinations, (MNIST, CIFAR10, CIFAR100) none of which work. I do not really see how this would have worked as no matter how this is framed, repeating the data (x) by the batch size generates the variable x_batch. 

If I am not mistaken, I think this will always result in a size mismatch (it may be possible to get it to work using 

I have tested my change and it seems to work, but the batch size parameter is not doing anything and the attack is run. Maybe the person who implemented this attack can provide some insight.

I also found a workaround without modifying the code, simply set batch size equal to 1

mcguires5:
If we do not want to mess with this at the moment I would recommend at least changing the default batch size to 1 that way it won't error when someone tries to run this (until they change the batch size).

beat-buesser:
Yes, that's correct, `ShadowAttack` only works for a single example. It then takes `batch_size` random samples of it, which is usually a quite large number and reaching memory limits. Therefore the input `x` is limited to a single sample. We should update the docstring to mention this specifically.

mcguires5:
Okay so the problem was that `batch_size` here refers to the number of random instances created from the originals, I was interpreting that incorrectly, running on a single instance I ran into the problem with running out of memory until I reduced `batch_size` to 128. 

The documentation says "batch_size (int) – The size of the training batch." where it may be more helpful to write something explaining it refers to the number of random samples for each instance created. It is probably also worth updating "x – An array with the original inputs." to mentioning something about recommending only using a single sample at a time.

Thanks for the clarification!

PR #510, opened 2020-07-24
Chaining preprocessor defenses

mzweilin:
# Description

We implement an efficient chaining mechanism for preprocessor-style defenses using the autograd of PyTorch, along with an example defense `SpatialSmoothingPytorch`.

The new chaining mechanism requires all defenses implemented accordingly in PyTorch, see example in `art/defences/preprocessor/spatial_smoothing_pytorch.py`.

However, the new mechanism does work with existing numpy-based defenses as a single defense without chaining. As the current chaining mechanism is wrong with back-propagation, the new mechanism doesn't break any existing working features.

The differentiable spatial smoothing depends on "kornia==0.3.1" which has been added in `setup.py` and `requirements.txt` accordingly.

Fixes https://github.com/Trusted-AI/adversarial-robustness-toolbox/issues/485

## Type of change

Please check all relevant options.

- [x] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Testing

Please describe the tests that you ran to verify your changes. Consider listing any relevant details of your test configuration.

- [x] `pytest -q -s tests/classifiersFrameworks/test_pytorch.py  --mlFramework=pytorch --durations=0`
- [x] `pytest -q -s tests/defences/preprocessor/test_spatial_smoothing_pytorch.py  --mlFramework=pytorch --durations=0`

**Test Configuration**:
- OS: Ubuntu 18.04.4 LTS
- Python version: 3.7.5
- ART version or commit number: 09262467e6f0a5832d2580b0bf995ecee333171d
- PyTorch 1.5.0

# Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #574, opened 2020-08-25
Extend framework-specific preprocessing defences to TensorFlow v2

beat-buesser:
# Description

This pull request extends the framework-specific defences introduced in #510 to support TensorFlow v2.

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/art/defences/preprocessor/preprocessor.py
+++ b/art/defences/preprocessor/preprocessor.py
@@ -21,9 +21,12 @@ This module implements the abstract base class for defences that pre-process inp
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import abc
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, TYPE_CHECKING
 
 import numpy as np
+
+if TYPE_CHECKING:
+    import torch
 
 
 class Preprocessor(abc.ABC):
--- a/art/defences/preprocessor/preprocessor.py
+++ b/art/defences/preprocessor/preprocessor.py
@@ -125,6 +125,8 @@ class PreprocessorPyTorch(Preprocessor):
     Abstract base class for preprocessing defences implemented in PyTorch that support efficient preprocessor-chaining.
     """
 
+    import torch
+
     @abc.abstractmethod
     def forward(self, x: torch.Tensor, y: Optional[torch.Tensor] = None) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
         """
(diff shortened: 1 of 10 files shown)


art/defences/preprocessor/preprocessor.py
```diff
@@ -21,10 +21,14 @@
...
 import abc
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, TYPE_CHECKING
 
 import numpy as np
 
+if TYPE_CHECKING:
+    import torch
+    import tensorflow as tf
+
```
minhitbk, line 31: I am not sure if when you import torch and tensorflow at the same time, it may make some confict on gpus if they are available becaust tf will occupy all gpus by default, and then torch doesn't have one to use. However, here is just a type check and maybe it won't effect but I think you should check.

art/estimators/tensorflow.py
```diff
@@ -119,3 +121,126 @@ def fit(self, x: np.ndarray, y, batch_size: int = 128, nb_epochs: int = 20, **kw
...
         NeuralNetworkMixin.fit(self, x, y, batch_size=128, nb_epochs=20, **kwargs)
+
+    def _apply_preprocessing_defences(self, x, y, fit: bool = False) -> Tuple[Any, Any]:
+        """
+        Apply all preprocessing defences of the estimator on the raw inputs `x` and `y`. This function is should
+        only be called from function `_apply_preprocessing`.
+
+        The method overrides art.estimators.estimator::BaseEstimator._apply_preprocessing_defences().
+        It requires all defenses to have a method `forward()`.
+        It converts numpy arrays to PyTorch tensors first, then chains a series of defenses by calling
```
minhitbk, line 132: I think this version is for tensorflow v2, so you need to update the document to tensorflow v2. I see many other places still keep pytorch in a tensorflow document.

minhitbk (approved):
Hi @beat-buesser,

This is a good work, it helps to make code clearner wrt the framework support. I just have minor comments on the documents of Tensorflow code where you haven't updated it from Pytorch to Tensorflow. Please update those and feel free to merge.

art/defences/preprocessor/preprocessor.py
beat-buesser: In general for productions, I agree with your concerns, but as you mention these imports will only be carried out during the type-checking with mypy and so far we have not observed any problems.

art/estimators/tensorflow.py
beat-buesser: Great catch! I have changed all `pytorch` references to `tensorflow` references.

beat-buesser:
Hi @minhitbk Thank you very much for the review! I have made all required updates.

Issue #273, opened 2020-01-28
Implement attack for imperceptible, robust and targeted adversarial examples for Automatic Speech Recognition


beat-buesser:
Implement attack for imperceptible, robust and targeted adversarial examples for Automatic Speech Recognition.

Paper link: https://arxiv.org/abs/1903.10346

- investigate possible reuse/extension of or similarities with `ExpectationOverTransformations`
- full integration of this attack most likely depends on #272

beat-buesser:
Reopening for Lingvo estimator and framework-independent ASR attack.

PR #584, opened 2020-09-02
Add Lingvo ASR estimator

hesseltuinhof:
# Description

This PR adds a model-specific Lingvo ASR estimator for sequence models. Such an estimator is required to implement the attack described in #273

## Type of change

# Testing

Several tests have been written for the estimator.

**Test Configuration**:
- Fedora 31
- Python 3.6.10
- ART 1.5 development
- Lingvo 0.6.4 with TensorFlow 2.1.0 

# Checklist


requirements.txt
```diff
@@ -27,13 +27,18 @@ GPy==1.9.9
 lightgbm==2.3.1
 xgboost==1.1.1
 
+# Lingvo ASR dependencies
+lingvo==0.7.2
```
beat-buesser, line 31: Could you please update Dockerfile accordingly?

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+        params_dir = os.path.join(self._LINGVO_CFG["path"], "asr")
+        params_base = "librispeech.py"
+        if not os.path.isdir(params_dir):
+            make_directory(params_dir)
+        if not os.path.isfile(os.path.join(params_dir, params_base)):
+            logger.info("Could not find %s. Downloading it now...", params_base)
+            get_file(params_base, self._LINGVO_CFG["params_uri"], path=params_dir)
+        return os.path.join(params_dir, params_base)
+
+    def _check_and_download_model(self) -> str:
```
beat-buesser, line 158:
```suggestion
    def _check_and_download_model(self) -> Tuple[str, str]:
```

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+        # extract log-mel features
+        log_mel = mel_frontend.FPropDefaultTheta(NestedMap(src_inputs=x, paddings=tf1.zeros_like(x)))
+        features = log_mel.src_inputs
+
+        # reshape features to shape (batch_size, n_frames, n_features, channels) in compliance with Qin et al.
+        features_shape = (tf1.shape(x)[0], -1, 80, features.shape[-1])
+        features = tf1.reshape(features, features_shape)
+        return features
+
+    def _pad_audio_input(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
```
beat-buesser, line 281:
```suggestion
    def _pad_audio_input(self, x: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+            }
+            # get loss gradient
+            gradient = self._sess.run(
+                self._loss_gradient(self._x_padded, self._y_target, self._mask_frequency), feed_dict
+            )
+            gradients.append(np.squeeze(gradient))
+
+        return np.array(gradients)
+
+    def set_learning_phase(self) -> NoReturn:
```
beat-buesser, line 449: The `NoReturn` leads to a `Signature of incompatible with supertype "NeuralNetworkMixin"` error in mypy. Is there a way to keep `NoReturn` for `NotImplementedError` cases but avoid the mpypy error?

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+        raise NotImplementedError
+
+    def get_activations(self) -> NoReturn:
```
beat-buesser, line 452: The `NoReturn` leads to a `Signature of incompatible with supertype "NeuralNetworkMixin"` error in mypy. Is there a way to keep `NoReturn` for `NotImplementedError` cases but avoid the mpypy error?

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+        model_ckpt_index_base = "ckpt-00908156.index"
+        if not os.path.isdir(model_ckpt_dir):
+            make_directory(model_ckpt_dir)
+        if not os.path.isfile(os.path.join(model_ckpt_dir, model_ckpt_data_base)):
+            logger.info("Could not find %s. Downloading it now...", model_ckpt_data_base)
+            get_file(model_ckpt_data_base, self._LINGVO_CFG["model_ckpt_data_uri"], path=model_ckpt_dir)
+        if not os.path.isfile(os.path.join(model_ckpt_dir, model_ckpt_index_base)):
+            logger.info("Could not find %s. Downloading it now...", model_ckpt_index_base)
+            get_file(model_ckpt_index_base, self._LINGVO_CFG["model_ckpt_index_uri"], path=model_ckpt_dir)
+        return (os.path.join(model_ckpt_dir, model_ckpt_data_base), os.path.join(model_ckpt_dir, model_ckpt_index_base))
```
beat-buesser, line 171:
```suggestion
        return os.path.join(model_ckpt_dir, model_ckpt_data_base), os.path.join(model_ckpt_dir, model_ckpt_index_base)
```

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+        else:
+            gradients = self._loss_gradient_per_sequence(x, y)
+        return gradients
+
+    def _loss_gradient_per_batch(self, x: np.ndarray, y: np.ndarray) -> np.ndarray:
+        """
+        Compute the gradient of the loss function w.r.t. `x` per batch.
+        """
+        import tensorflow.compat.v1 as tf1
+        from lingvo.core.py_utils import NestedMap
```
beat-buesser, line 398:
```suggestion
```

art/estimators/sequence/tensorflow.py
```diff
@@ -0,0 +1,453 @@
...
+            gradients.append(gradient)
+
+        return np.array(gradients)
+
+    def _loss_gradient_per_sequence(self, x: np.ndarray, y: np.ndarray) -> np.ndarray:
+        """
+        Compute the gradient of the loss function w.r.t. `x` per sequence.
+        """
+        import tensorflow.compat.v1 as tf1
+        from lingvo.core.py_utils import NestedMap
```
beat-buesser, line 429:
```suggestion
```

beat-buesser (changes requested):
Hi @hesseltuinhof Thank you very much for your pull request! This code will greatly facilitate working with Lingvo ASR! I have a few request and suggestions, but otherwise the code looks good.

art/estimators/sequence/tensorflow.py
hesseltuinhof: Thanks for highlighting the `mypy` error. I overlooked that initially.

I think we should use a `Union` in the return type of the `NeuralNetworkMixin`. I have added a note to pick this up myself later. Reverted the type now back to the return type of the abstract method definition (which are technically the wrong types if we raise the `NotImplementedError`...).

art/estimators/sequence/tensorflow.py
hesseltuinhof: See previous.

hesseltuinhof:
@beat-buesser, thanks for catching some small problems and helping to improve the PR, which I updated accordingly. 

Please note, that I moved the Lingvo estimator into the same module of the recent `Deepspeech` estimator addition of @minhitbk.

hesseltuinhof:
Recent work on the estimator, adds GPU support for the Lingvo estimator! Unfortunately, this comes at the price of pinning Lingvo to version `0.6.4` and TensorFlow to version `2.1.0`. As a result, setting up a development environment via the `requirements.txt` or `Dockerfile` requires manual intervention, namely one needs to uncomment the lines related to Lingvo and comment out some other TF dependencies.

However, the user can install ART with Lingvo support without any manual intervention by just calling `pip install adversarial-robustness-toolbox[lingvo_asr]` (or from a development branch with `pip install -e .[lingvo_asr]`).

One open issue remains, namely how to enable CI to run tests for the Lingvo estimator? Currently, they are disabled as our Python 3.6 job runs TensoFlow version `1.15.3`... :thinking:

beat-buesser:
Hi @hesseltuinhof, thank you very much for creating an estimator for Lingvo and making it work on GPUs. I think this new Lingvo estimator will soon find a lot of applications. Because of the specific dependency requirements I would suggest to add an additional Travis run for this version combination required by the Lingvo estimator, but only run the Lingvo tests with this new Travis run.

hesseltuinhof:
@beat-buesser and @killianlevacher the PR is now ready for review :100: 

Please note that related tests are skipped on CI because CI doesn't support a `tensorflow==2.1.0` enviroment. As discussed earlier, I will add an additional Github Actions runner for the required environment. However, in an seperate PR.

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+    sample_rate = 16000
+    frequency_length = (sample_rate // 2 + 1) // 240 * 3
+    test_input = np.zeros((2, sample_rate))
+    test_mask_frequency = np.ones((2, frequency_length, 80))
+    return test_input, test_mask_frequency
+
+
+@pytest.mark.skipif(tf.__version__ != "2.1.0", reason="requires Tensorflow 2.1.0")
+@pytest.mark.skipMlFramework("pytorch", "mxnet", "kerastf", "non_dl_frameworks")
+class TestLingvoAsr:
```
killianlevacher, line 68: The class should be removed and markers placed in front of each test instead. See section 3.1.3 in [wiki doc](https://github.com/Trusted-AI/adversarial-robustness-toolbox/wiki/ART-Unit-Testing) and example class [test_deeplearning_common.py](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/tests/estimators/classification/test_deeplearning_common.py)

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+    """
+    Test the LingvoAsr estimator.
+    """
+
+    def test_is_subclass(self):
```
killianlevacher, line 73: In the new test framework, for each method we need to add a default `try / except ARTTestException as e: art_warning(e)` at the beginning and end of the test. This is a safeguard which guarantees any ART Meta test Warnings are brought to our attention even if the test passes. See section 2 in [wiki doc](https://github.com/Trusted-AI/adversarial-robustness-toolbox/wiki/ART-Unit-Testing) and example class [test_deeplearning_common.py](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/tests/estimators/classification/test_deeplearning_common.py)

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+
+from art.estimators.speech_recognition.speech_recognizer import SpeechRecognizerMixin
+from art.estimators.speech_recognition.tensorflow_lingvo import LingvoAsr
+from art.estimators.tensorflow import TensorFlowV2Estimator
+from art.utils import get_file
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
```
killianlevacher, line 36: 1/ So one of the conventions we are using in the new tests is to place any new fixtures within `conftest.py` files in the relevant package. This is to make sure we reuse as much as possible fixtures across test scripts to avoid duplication of code. So in this case both `audio_data` and `audio_batch_padded` fixtures should be moved to a new `conftest.py` file located at `tests/estimators/speech_recognition/`. Please see See section 3.2 in [wiki doc](https://github.com/Trusted-AI/adversarial-robustness-toolbox/wiki/ART-Unit-Testing) about creating fixtures. 

Also, I guess we can move this to Slack but I'll also make sure that these fixtures are added to the wiki test documentation so that other people can reuse that fixture in the future if needed.

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+class TestLingvoAsr:
+    """
+    Test the LingvoAsr estimator.
+    """
+
+    def test_is_subclass(self):
+        assert issubclass(LingvoAsr, (SpeechRecognizerMixin, TensorFlowV2Estimator))
+
+    def test_implements_abstract_methods(self):
+        import tensorflow.compat.v1 as tf1
```
killianlevacher, line 77: in the same spirit as removing duplicated code, line `import tensorflow.compat.v1 as tf1` and `tf1.reset_default_graph()` should be removed from all the tests and instead placed in the `def setup_tear_down_framework(framework):` fixture in the project's main `conftest.py` file. As you'll see, this is already done for TF2 so it's a nice addition for TF1 :-)

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+            lingvo._x_padded: test_input,
+            lingvo._y_target: test_target_dummy,
+            lingvo._mask_frequency: test_mask_frequency,
+        }
+        loss_gradient = lingvo._sess.run(lingvo._loss_gradient_op, feed_dict)
+        assert test_input.shape == loss_gradient.shape
+        assert loss_gradient.sum() == 0.0
+
+    def test_loss_gradient_per_batch(self, audio_data):
+        import tensorflow.compat.v1 as tf1
```
killianlevacher, line 204: aren't `test_loss_gradient_per_batch` and `test_loss_gradient_per_sequence` duplicated tests? I'm trying to see how do they differ but can't see any difference. 
In general, there's 2 conventions that would apply here. 
Option 1: If the only thing that differs from the tests is an input parameter, then both tests should be merged into one and instead the `@pytest.mark.parametrize` should be used (see [test_fast_gradient.py/test_norm_images](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/tests/attacks/evasion/test_fast_gradient.py) as an example)

Option 2: if there's many differences between the two but still quite a lot of code duplicated between them then we need to refactor the identical code into a separate method called by both tests named `back_end_my_function_etc` (see section 3.3.3 in [wiki doc](https://github.com/Trusted-AI/adversarial-robustness-toolbox/wiki/ART-Unit-Testing))

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # test shape, equal inputs have equal gradients, non-zero inputs have non-zero gradient sums
+        assert [x.shape for x in test_input] == [g.shape for g in gradients]
+        assert_allclose(np.abs(gradients[2]).sum(), np.abs(gradients[3]).sum(), rtol=1e-01)
+        assert_array_equal(gradients_abs_sum > 0, [False, True, True, True])
+
+
+@pytest.mark.skipif(tf.__version__ != "2.1.0", reason="requires Tensorflow 2.1.0")
+@pytest.mark.skipMlFramework("pytorch", "mxnet", "kerastf", "non_dl_frameworks")
+class TestLingvoAsrLibriSpeechSamples:
```
killianlevacher, line 242: same thing as above with respect to the class

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+            ),
+        },
+    }
+
+    def test_predict(self):
+        import tensorflow.compat.v1 as tf1
+
+        tf1.reset_default_graph()
+
+        transcripts = list()
```
killianlevacher, line 282: similarly to a previous comment about duplicated code, in this case this is a clear nice use case of using option 2 and creating a `back_end_method` with lines from 282 to 292 in it. This method can then be called by both `test_predict` and `test_loss_gradient` tests (see section 3.3.3 in [wiki doc](https://github.com/Trusted-AI/adversarial-robustness-toolbox/wiki/ART-Unit-Testing))

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # test loss gradients per batch and per sequence are the same
+        assert_allclose(gradient_sequence_sum, gradient_batch_sum, rtol=1e-05)
+        # test gradient_batch, gradient_sequence and audios items have same shapes
+        assert (
+            [gb.shape for gb in gradient_batch] == [gs.shape for gs in gradient_sequence] == [a.shape for a in audios]
+        )
+
+
+if __name__ == "__main__":
```
killianlevacher, line 328: this should be removed

run_tests.sh
```diff
@@ -23,6 +23,9 @@ for mlFramework in "${mlFrameworkList[@]}"; do
   pytest --cov-report=xml --cov=art --cov-append  -q -vv -s tests/attacks/evasion/ --mlFramework=$mlFramework  --skip_travis=True --durations=0
   if [[ $? -ne 0 ]]; then exit_code=1; echo "Failed attacks/evasion/test_shadow_attack.py"; fi
 
+  pytest -q tests/estimators/speech_recognition/ --mlFramework=$mlFramework --durations=0
```
killianlevacher, line 26:
```suggestion
  pytest  --cov-report=xml --cov=art --cov-append  -q tests/estimators/speech_recognition/ --mlFramework=$mlFramework --skip_travis=True --durations=0
```

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
killianlevacher, line 328:
```suggestion
```

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+    pytest.cmdline.main("-q -s {} --mlFramework=tensorflow --durations=0".format(__file__).split(" "))
```
killianlevacher, line 329:
```suggestion
```

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
```diff
@@ -0,0 +1,329 @@
...
+        assert [x.shape for x in test_input] == [g.shape for g in gradients]
+        assert_allclose(np.abs(gradients[2]).sum(), np.abs(gradients[3]).sum(), rtol=1e-01)
+        assert_array_equal(gradients_abs_sum > 0, [False, True, True, True])
+
+
+@pytest.mark.skipif(tf.__version__ != "2.1.0", reason="requires Tensorflow 2.1.0")
+@pytest.mark.skipMlFramework("pytorch", "mxnet", "kerastf", "non_dl_frameworks")
+class TestLingvoAsrLibriSpeechSamples:
+    # specify LibriSpeech samples for download and with transcriptions
+    samples = {
```
killianlevacher, line 244: With the new test framework, we have a new nice functionality which enables us to hide long arrays or json from the test code to make it more readable. This is a perfect use case with the variable `samples`. By using the `store_expected_values` fixture to store that json object and then the `expected_values` fixture in each test to retrieve it, this will automatically create a separate json file for this test script which each test will leverage. There's more detailed explanation in section 6 of the [wiki](https://github.com/Trusted-AI/adversarial-robustness-toolbox/wiki/ART-Unit-Testing). It really just takes a few seconds to do but if anything is unclear, I'm more than happy to help you of course :-)

hesseltuinhof: Good point. Maybe we should go even one step further and add the three Librispeech samples as fixtures to the tests, e.g. similar as we do for images with MNIST data. That would also make your previous comment redundant regarding the duplicate code for loading the samples. 

I would suggest to add this as a seperate feature request for the near future.

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
hesseltuinhof: See comment related to the `samples` dictionary.

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+    from art.defences.preprocessor.preprocessor import Preprocessor
+    from art.defences.postprocessor.postprocessor import Postprocessor
+
+    from tensorflow.compat.v1 import Tensor
+    from tensorflow.compat.v1 import Session
+
+logger = logging.getLogger(__name__)
+
+
+class LingvoAsr(SpeechRecognizerMixin, TensorFlowV2Estimator):
```
beat-buesser, line 46: How about this class name, it would be closer to the `art.estimators` pattern:
```suggestion
class TensorFlowLingvoASR(SpeechRecognizerMixin, TensorFlowV2Estimator):
```

art/estimators/speech_recognition/__init__.py
```diff
@@ -4,3 +4,4 @@
 from art.estimators.speech_recognition.speech_recognizer import SpeechRecognizerMixin
 
 from art.estimators.speech_recognition.pytorch_deep_speech import PyTorchDeepSpeech
+from art.estimators.speech_recognition.tensorflow_lingvo import LingvoAsr
```
beat-buesser, line 7:
```suggestion
from art.estimators.speech_recognition.tensorflow_lingvo import TensorFlowLingvoASR
```

Dockerfile
```diff
@@ -13,6 +13,13 @@ RUN pip3 install pandas==1.1.1
 RUN pip3 install jupyter==1.0.0 && pip3 install jupyterlab==2.1.0
 # https://stackoverflow.com/questions/49024624/how-to-dockerize-jupyter-lab
 
+# Lingvo ASR dependencies
+# supported versions: (lingvo==0.6.4 with tensorflow==2.1.0)
```
beat-buesser, line 17: I have noticed that `pip install lingvo==0.6.4` is installing `tensorflow-gpu` as a dependency even if `tensorflow` is installed. Do you know if there is a way to install TensorFlow only once if we install ART with the new option `lingvo_asr`?

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
hesseltuinhof: See above.

requirements.txt
```diff
@@ -29,6 +29,12 @@ lightgbm==3.0.0
 xgboost==1.2.0
 kornia==0.4.0
 
+# Lingvo ASR dependencies
+# supported versions: (lingvo==0.6.4 with tensorflow==2.1.0)
```
beat-buesser, line 33: Same comment as on Dockerfile: Should we update this comment to `tensorflow-gpu` to only install a single TensorFlow?

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+        if self.preprocessing_defences is not None:
+            raise ValueError("This estimator does not support `preprocessing_defences`.")
+        if self.postprocessing_defences is not None:
+            raise ValueError("This estimator does not support `postprocessing_defences`.")
+
+        # check required TensorFlow version
+        assert tf1.__version__ == "2.1.0", "The Lingvo estimator only supports TensorFlow 2.1.0."
+
+        # check required Python version
+        assert sys.version_info[:2] == (3, 6), "The Lingvo estimator only supports Python 3.6."
```
beat-buesser, line 157: Should we transform these checks into `if` and `raise` exceptions to avoid users running into unexpected behavior with `-O` flag?

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+
+        # check required Lingvo version
+        assert (
+            pkg_resources.get_distribution("lingvo").version == "0.6.4"
+        ), "The Lingvo estimator only supports Lingvo 0.6.4"
```
beat-buesser, line 162: Should we transform these checks into `if` and `raise` exceptions to avoid users running into unexpected behavior with `-O` flag?

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
hesseltuinhof: I understand that we want to bound markers to the method. Therefore, I would suggest to adapt the documentation to explicityly ask for this. But to continue to allowing test classes. Even if we disregard the benefits of using test classes for sharing fixtures or apply marks at the class level, we still allow the benefit to group multiple tests and increase code organization. Also, note that `pytest` will ensure that each test inside a test class has it's own class instance.

What are your opinions @killianlevacher and @beat-buesser ?

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+    ) -> Union[Tuple[np.ndarray, np.ndarray], np.ndarray]:
+        """
+        Perform batch-wise prediction for given inputs.
+
+        :param x: Samples of shape `(nb_samples)` with values in range `[-32768, 32767]`. Note that it is allowable
+                  that sequences in the batch could have different lengths. A possible example of `x` could be:
+                  `x = np.ndarray([[0.1, 0.2, 0.1, 0.4], [0.3, 0.1]])`.
+        :param batch_size: Size of batches.
+        :return: Array of predicted transcriptions of shape `(nb_samples)`. A possible example of a transcription
+                 return is `np.array(['SIXTY ONE', 'HELLO'])`.
```
beat-buesser, line 392: It seems that `predict` return all lowercase letters whereas the docstring and `PyTorchDeepSpeech` return all uppercase letters. Should we change the output to uppercase?

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+        """
+        if x[0].ndim != 1:
+            raise ValueError(
+                "The LingvoAsr estimator can only be used temporal data of type mono. Please remove any channel"
+                "dimension."
+            )
+
```
beat-buesser, line 399: We need to apply the preprocessing operations:
```suggestion

        # Apply preprocessing
        x_preprocessed, _ = self._apply_preprocessing(x_, y=None, fit=False)

```

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+        :param y: Target values of shape (nb_samples). Each sample in `y` is a string and it may possess different
+                  lengths. A possible example of `y` could be: `y = np.array(['SIXTY ONE', 'HELLO'])`.
+        :param batch_mode: If `True` calculate gradient per batch or otherwise per sequence.
+        :return: Loss gradients of the same shape as `x`.
+        """
+        if batch_mode:
+            gradients = self._loss_gradient_per_batch(x, y)
+        else:
+            gradients = self._loss_gradient_per_sequence(x, y)
+        return gradients
```
beat-buesser, line 458: We also need preprocessing for loss gradients, maybe something like this:
```suggestion
        """
        
        # Apply preprocessing
        x_preprocessed, y_preprocessed = self._apply_preprocessing(x, y, fit=False)
        
        if batch_mode:
            gradients = self._loss_gradient_per_batch(x_preprocessed, y_preprocessed)
        else:
            gradients = self._loss_gradient_per_sequence(x_preprocessed, y_preprocessed)
           
        gradients = self._apply_preprocessing_gradient(x, gradients) 
           
        return gradients
```

Dockerfile
hesseltuinhof: Yes, this is a typo in the comment :+1:

art/estimators/speech_recognition/tensorflow_lingvo.py
beat-buesser, line 458: I'm not completely sure, but it looks like for `x` with dtype `np.object` this `loss_gradient` returns a single array which results in shape checks failing in `ProjectedGradientDescent` and `FastGradientMethod` if `y=None` in their method `generate`. Can we change the output type and shape to be the same as `x` for arrays and for arrays of arrays?

art/estimators/speech_recognition/tensorflow_lingvo.py
```diff
@@ -0,0 +1,516 @@
...
+
+        :param clip_values: Tuple of the form `(min, max)` of floats or `np.ndarray` representing the minimum and
+               maximum values allowed for features. If floats are provided, these will be used as the range of all
+               features. If arrays are provided, each value will be considered the bound for a feature, thus
+               the shape of clip values needs to match the total number of features.
+        :param channels_first: Set channels first or last.
+        :param preprocessing_defences: Preprocessing defence(s) to be applied by the classifier.
+        :param postprocessing_defences: Postprocessing defence(s) to be applied by the classifier.
+        :param preprocessing: Tuple of the form `(subtrahend, divisor)` of floats or `np.ndarray` of values to be
+            used for data preprocessing. The first value will be subtracted from the input. The input will then
```
beat-buesser, line 129: Please update indentation.

art/estimators/speech_recognition/tensorflow_lingvo.py
hesseltuinhof: Yes, this makes sense in order to be consistent with the `PyTorchDeepspeech` estimator.

beat-buesser (changes requested):
Hi @hesseltuinhof thank you very much for this pull request, this estimator will be very useful for many users of ART! Code looks great. I have made a few change request to facilitate integration into ART and based on experiments with `ProjectedGradientDescent`.

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
killianlevacher: sure, yes that's a good idea. The most important really is that the markers are all set at the method level and not at the class level. This is to ensure that there is no hierarchical dependency on the markers that regulate which frameworks should be used for each test method. Also as you mentioned, we really need to be careful about not putting setup/teardown methods in these classes then an instead reuse ART fixtures as much as possible. I will update the wiki documentation to reflect this clarification. Thanks for your suggestion.

tests/estimators/speech_recognition/test_tensorflow_lingvo.py
killianlevacher: yes I agree that would be very good indeed, as I'm sure we will be reusing these audio files as ART further develops its speech functionalities. Also it would make the test more robust as it would not be dependent on downloading this file for each test. @beat-buesser what's the procedure for this, can Hessel create a new issue for this so that we can track the need for this feature?

hesseltuinhof: I have created an issue for adding Librispeech sample fixtures :+1:

art/estimators/speech_recognition/tensorflow_lingvo.py
hesseltuinhof: I couldn't reproduce this. This probably was related to the wrong batch size in your test cases, which initially raised exception. I am resolving the converation.

beat-buesser (approved):

beat-buesser:
@hesseltuinhof Thank you very much for this great contribution, I'm sure it will soon be used by many users of ART!

PR #678, opened 2020-10-20
Update Spectral Signatures Defense

### LINKED ISSUE #593: Investigate bad performance in Spectrum Signatures poisoning defense
**Describe the bug**
Defense performance is not as good as expected. 

**To Reproduce**
Unexpected performance with custom dataset. 

**Expected behavior**
Better performance is expected in terms of detection rates.

**Screenshots**

 - Mac OS or Ubuntu both tested 
 - Python version: 3.7 
 - ART version or commit number: ART 1.3 release
 -  TF 1.15

ebubae:
# Description

This PR:

- Adds type safety to `PoisonFilteringDefence` abstract class
- removes `num_classes` param of `SpectralSignatureDefence`
- rename `ub_pct_poison` to `expected_pp_poison` to enhance readability
- increases re-use of ART utility methods
- handles edge cases for defense input (no training points for class)
- adds a notebook for `SpectralSignatureDefence`
- updates tests to modern ART testing conventions
- updates types to be more liberal for certain utility functions
- fixes a bug in a the spectral score calculation

Resolves #593 

## Type of change

- [x] Bug fix (non-breaking)

Removes `num_classes` parameter for defense. This is now automatically determined based on the classifier.
This PR also replaces to `ub_pct_poison` hyper-parameter to be named `expected_pp_poison`.

# Testing

Updated tests to new ART testing format

# Checklist


art/utils.py
```diff
@@ -1096,7 +1096,7 @@ def preprocess(
     return normalized_x, categorical_y
 
 
-def segment_by_class(data: np.ndarray, classes: np.ndarray, num_classes: int) -> List[np.ndarray]:
+def segment_by_class(data: Union[np.ndarray, List[int]], classes: np.ndarray, num_classes: int) -> List[np.ndarray]:
```
beat-buesser, line 1099: Why do is list of integer type required? Would it be possible to convert the list input into a ndarray?

art/defences/detector/poison/ground_truth_evaluator.py
```diff
@@ -40,7 +40,7 @@ def __init__(self):
         """
 
     def analyze_correctness(
-        self, assigned_clean_by_class: np.ndarray, is_clean_by_class: list
+        self, assigned_clean_by_class: Union[np.ndarray, List[np.ndarray]], is_clean_by_class: list
```
beat-buesser, line 43: Why is the list of np.ndarray type required?

art/defences/detector/poison/spectral_signature_defense.py
```diff
@@ -139,38 +141,28 @@ def detect_poison(self, **kwargs) -> Tuple[dict, List[int]]:
...
         if self.eps_multiplier < 0:
             raise ValueError("eps_multiplier must be positive. Unsupported value: " + str(self.eps_multiplier))
-        if self.ub_pct_poison < 0 or self.ub_pct_poison > 1:
-            raise ValueError("ub_pct_poison must be between 0 and 1. Unsupported value: " + str(self.ub_pct_poison))
+        if self.expected_pp_poison < 0 or self.expected_pp_poison > 1:
+            raise ValueError("expected_pp_poison must be between 0 and 1. Unsupported value: " +
+                             str(self.expected_pp_poison))
+
+
+def spectral_signature_scores(matrix_r: np.ndarray) -> np.ndarray:
```
beat-buesser, line 157: Should we make this function a static method of `SpectralSignatureDefense`?

tests/defences/detector/poison/test_spectral_signature_defense.py
```diff
@@ -18,72 +18,61 @@
...
-            batch_size=BATCH_SIZE,
-            eps_multiplier=EPS_MULTIPLIER,
-            ub_pct_poison=UB_PCT_POISON,
-            nb_classes=10,
-        )
 
-    def setUp(self):
-        # Set master seed
-        master_seed(1234)
+@pytest.mark.xfail
```
beat-buesser, line 34: Why is this test expected to fail?

beat-buesser (changes requested):
Hi @ebubae Thank you very much for the updates! I have a few question about the changes.

Would it be possible to hide the TensorFlow warnings in cell 6 of the new notebook?

art/defences/detector/poison/ground_truth_evaluator.py
ebubae: The `Union` operator allows either of the types inside. This change allows both np arrays and lists of np arrays to be accepted. This is helpful because the output of another util method, `segment_by_class`, returns a list of np arrays.

Because of the way the function enumerates through the `assigned_clean_by_class` parameter, the output is actually the same for lists of np arrays and np arrays formatted correctly from lists. This allows us to remove an extra step of processing by using the lists as is.

tests/defences/detector/poison/test_spectral_signature_defense.py
ebubae: The params that are passed in as an argument are created by the `parameterize` method above. All of those params are invalid so this check is to ensure that invalid params are checked by the defense.

art/utils.py
ebubae: The `Union` operator won't require that the input is a list, but it will allow `List[int]` as well as `np.ndarray` in this function.

Here is how the `data` parameter is used in this function:
```python
by_class[assigned].append(data[indx])
```

The way the `data` argument is used in this function operates the same for `ndarray` and lists. Given this is a convenience utility method I think it's okay to accept both if that suits the use case.

beat-buesser (approved):

PR #685, opened 2020-10-24
Update AutoProjectedGradientDescent

beat-buesser:

# Description

This pull request provides updates for `AutoProjectedGradientDescent`. It extends `AutoProjectedGradientDescent` to run with estimators of all frameworks using the estimator's loss function requiring the estimator to predict logits and provide loss gradients. It updates `AutoProjectedGradientDescent` to use the new method `loss` of the Estimator API replacing custom functions in `AutoProjectedGradientDescent` and implements method `loss` for `TensorFlowClassifer`. It disables the loss type `difference_logits_ratio` for `TensorFlowClassifer`, for now until a suspected bug in that loss function has been fixed, by commenting the implementation as it seems to fail to select the second largest logit for cases where the largest logit is the true logit which occurs in small number of predictions, this only affects `AutoProjectedGradientDescent` with `TensorFlowClassifer`.)

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/art/attacks/evasion/auto_projected_gradient_descent.py
+++ b/art/attacks/evasion/auto_projected_gradient_descent.py
@@ -89,258 +89,250 @@ class AutoProjectedGradientDescent(EvasionAttack):
         """
         from art.estimators.classification import TensorFlowClassifier, TensorFlowV2Classifier, PyTorchClassifier
 
-        if isinstance(estimator, TensorFlowClassifier):
-            import tensorflow as tf
-
-            if loss_type == "cross_entropy":
-                if is_probability(estimator.predict(x=np.ones(shape=(1, *estimator.input_shape)))):
-                    raise NotImplementedError("Cross-entropy loss is not implemented for probability output.")
-                else:
-                    self._loss_object = tf.reduce_mean(
-                        tf.keras.losses.categorical_crossentropy(
-                            y_pred=estimator._output, y_true=estimator._labels_ph, from_logits=True
-                        )
-                    )
-
-                    def loss_fn(y_true, y_pred):
-                        y_pred_norm = y_pred - np.amax(y_pred, axis=1, keepdims=True)
-                        loss_value = -(
-                            y_true * y_pred_norm - np.log(np.sum(np.exp(y_pred_norm), axis=1, keepdims=True))
-                        )
-                        return np.mean(loss_value)
-
-                    self._loss_fn = loss_fn
-            elif loss_type == "difference_logits_ratio":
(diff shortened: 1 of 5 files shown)


art/estimators/classification/keras.py
```diff
@@ -300,14 +300,18 @@ def _initialize_params(
         # Get the internal layer
         self._layer_names = self._get_layers()
 
-    def loss(self, x: np.ndarray, y: np.ndarray, **kwargs) -> np.ndarray:
+    def loss(self, x: np.ndarray, y: np.ndarray, reduction: str = "none", **kwargs) -> np.ndarray:
```
hesseltuinhof, line 303: Have you also considered using `reduction: Optional[str] = None` here?

art/attacks/evasion/auto_projected_gradient_descent.py
```diff
@@ -89,258 +89,255 @@ def __init__(
...
+                        #     z_i = np.diag(z_i)
+                        #     z_y = np.diag(z_y)
+                        #
+                        #     dlr = -(z_y - z_i) / (z_1 - z_3)
+                        #
+                        #     return np.mean(dlr)
+                        #
+                        # self._loss_fn = loss_fn
+                        # self._loss_object = difference_logits_ratio(y_true=estimator._labels_ph,
+                        #                                             y_pred=estimator._output)
```
hesseltuinhof, line 182: Looks like this code is replaced by the two `difference_logits_ratio` classes. Can we remove these comments here?

hesseltuinhof (changes requested):
I have added two comments. Otherwise, the PR looks ready for merging.

art/estimators/classification/keras.py
beat-buesser: Yes, I considered it. The reason for `"none"` and the options for `reductions was that it is what PyTorch is using for its loss functions.

art/attacks/evasion/auto_projected_gradient_descent.py
beat-buesser: Not completely, the other two `difference_logits_ratio` classes are specific for PyTorch and TensorFlow v2. The commented code here would be the class specific for TensorFlow v1, but it seems to contain a small bug as described in the exception above it and we keep the code commented until it is fix.

beat-buesser:
Hi @hesseltuinhof Thank you very much for your review, I have responded to your comments above.

art/estimators/classification/keras.py
hesseltuinhof: That's a good argument 👍

hesseltuinhof (approved):

PR #730, opened 2020-11-11
Framework Specific Github Actions Runs

### LINKED ISSUE #729: Github Action per Framework
We should separate the github action per frameworks. This would imply also keeping track of what tests are ran in each action to make sure each test is ran at least once

**Benefits**
This would give us the ability to retrieve tests results much faster (rather than having to wait for all the frameworks from being run) as well as reducing drastically the logs which will make them easier to read.
And more importantly, this would also enable us to test specific set up scenarios.
Doing so will also enable us to remove the need for the @frameworkAgnostic marker which was needed when using Travis. This would simplify even further writing tests, by removing the need to consider adding this marker or not

ebubae:
# Description

This PR changes the existing CI workflow in Github Actions to run each test framework separately in a matrix build. It also modifies `./run_tests.sh` to accept an argument indicating the ML framework to run (including `legacy` to run legacy tests)

Resolves #729 

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,12 @@ jobs:
           - framework: tensorflow
             python: 3.6
             tensorflow: 1.15.3
+            tf_version: v1
             keras: 2.2.5
           - framework: tensorflow
             python: 3.6
             tensorflow: 2.2.0
+            tf_version: v2
             keras: 2.3.1
           - framework: keras
             python: 3.7
@@ -52,7 +54,7 @@ jobs:
             tensorflow: 2.2.0
             keras: 2.3.1
 
-    name: Run ${{ matrix.framework }} ${{ matrix.framework == 'tensorflow' &&  matrix.tensorflow }} Tests
+    name: Run ${{ matrix.framework }}${{ matrix.tf_version }} Tests
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
@@ -71,7 +73,7 @@ jobs:
           python -m pip install --upgrade pip setuptools wheel
           pip3 install -q -r requirements.txt
           pip list
-      - name: Run ${{ matrix.framework }} ${{ matrix.framework == 'tensorflow' &&  matrix.tensorflow }} Tests
+      - name: Run ${{ matrix.framework }}${{ matrix.tf_version }} Tests
         run: ./run_tests.sh ${{ matrix.framework }}
   style:
     name: Style Check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
             tensorflow: 2.2.0
             keras: 2.3.1
 
-    name: Run ${{ matrix.framework }}${{ matrix.tf_version }} Tests
+    name: Run ${{ matrix.framework }} ${{ matrix.tf_version }} Tests
     steps:
       - name: Checkout Repo
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -25,13 +25,30 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - python: 3.6
+          - framework: tensorflow
```
killianlevacher, line 28:
```suggestion
          - framework: tensorflow1
```

.github/workflows/ci.yml
```diff
@@ -25,13 +25,30 @@ jobs:
...
+            python: 3.7
+            tensorflow: 2.2.0
+            keras: 2.3.1
+          - framework: pytorch
+            python: 3.7
+          - framework: scikitlearn
+            python: 3.7
+          - framework: mxnet
+            python: 3.7
+          - framework: legacy
```
killianlevacher, line 46:
```suggestion
          - framework: tensorflow2
```

.github/workflows/ci.yml
killianlevacher: I'm wondering whether there is any way of naming each matrix element so that we can distinguish them in the run names as` Run tensorflow1 Tests`(Instead of ` Run legacy Tests`) and `Run tensorflow2 Tests`? Both matrix elements should have the framework variable set to `tensorflow`.

run_tests.sh
```diff
@@ -9,160 +9,161 @@ export TF_CPP_MIN_LOG_LEVEL="3"
...
-  done
-done
-
-#bash <(curl -s https://codecov.io/bash)
-
-
-exit ${exit_code}
+mlFramework=$1
+
+if [[ ${mlFramework} != "legacy" ]]
```
killianlevacher, line 14: Due to the fact that we are in the process of transitioning the tests to the new system, it would be safer if we removed that if/else statement and left as it was before. This also relates to the previous comment about the fact that bot the `tensorflow1` and `tensorflow2` runs should both have the mlFramework variable set to `tensorflow`. The reason is because the pytest code within the `conftest.py` file takes care of determining internally whether `tensorflow` means `tensorflow1` or `tensorflow2` once the tests have started. Since this has no impact upon the legacy tests, it's safe to leave the run_tests.sh as it was before. In other words, unless I am mistaken only line 12 here is necessary.

run_tests.sh
```diff
@@ -9,160 +9,161 @@ export TF_CPP_MIN_LOG_LEVEL="3"
...
+      tests="$tests_module[@]"
+      for test in "${!tests}"; do
+         run_test ${test}
+      done
+    done
+
+    #bash <(curl -s https://codecov.io/bash)
+
+
+    exit ${exit_code}
```
killianlevacher, line 168: line 168 and 165 should also be outside of this else statement. This will be solved if the `if/else` statement is removed as mentioned in the previous comment above.

.github/workflows/ci.yml
killianlevacher, line 46: This is perfect. The only thing is that we need an extra matrix element identical to the `legacy` one but instead having the framework value set to `tensorflow`. This will make sure that the new non legacy tests, are also ran using tensorflow2

run_tests.sh
killianlevacher, line 168: I was initially going to suggest that we should remove the if/else statement but now I understand what you were trying to do and agree with this approach. The only small change I would make is to make sure that line 165 and 168 are outside of the if/else statement as we do want the script to produce the exit code regardless of whether the new or legacy tests were ran

.github/workflows/ci.yml
ebubae: Just added an extra `tf_version` matrix argument to do just this.

run_tests.sh
ebubae: Good catch. Fixed this now (but may have some caused MXNet errors).

beat-buesser (approved):
Hi @ebubae and @killianlevacher Thank you very much for this pull request! It greatly reduces the test time through parallelization!

Issue #587, opened 2020-09-04
Less strict dependencies


illeatmyhat:
All Trusted-AI libraries should have roughly the same version requirements.
As it is, they are incompatible with each other and you can't run them in the same environment if you have a strict dependency checker.

For that matter, a library should not lock to a specific version number, but should offer a range that still works.
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/925b6957870f8b689e24cac482f8b1c2afcc69d3/setup.py#L11-L13
The latest version of scikit-learn (0.23.2) is not even supported as a WML deployment.

i.e.
```python

 "scipy>=1.4.1", 
 "scikit-learn>=0.20.3",
```

beat-buesser:
Hi @illeatmyhat,

Thank you very much for the suggestion, which makes sense. We'll investigate how much we can relax the constraints for the upcoming release of ART 1.4. For most dependencies we don't have any strict requirements, but there are reasons for restrictions on `scipy` and `scikit-learn`.

ART is now a project of the Linux Foundation for AI and has therefore to be compatible with multiple deployment environments.

PR #615, opened 2020-09-22
Github Actions CI Pipeline

ebubae:
# Description

This PR adds a workflow for running our existing Travis tests using Github Actions. As an open source repository, we have much less limitations running tests in this environment. 

## Type of change

Please check all relevant options.

- [ ] Improvement (non-breaking)
- [ ] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #740, opened 2020-11-17
Extend and update test matrix

### LINKED ISSUE #734: Support for Python > 3.7 and/or scikit-learn > 0.22
I get the following error when I try to install with pip:

```
ERROR: Could not find a version that satisfies the requirement scikit-learn==0.22.2 (from adversarial-robustness-toolbox==1.4.2) (from versions: 0.9, 0.10, 0.11, 0.12, 0.12.1, 0.13, 0.13.1, 0.14, 0.14.1, 0.15.0b1, 0.15.0b2, 0.15.0, 0.15.1, 0.15.2, 0.16b1, 0.16.0, 0.16.1, 0.17b1, 0.17, 0.17.1, 0.18rc2, 0.18, 0.18.1, 0.18.2, 0.19b2, 0.19.0, 0.19.1, 0.19.2, 0.20rc1, 0.20.0, 0.20.1, 0.20.2, 0.20.3, 0.20.4, 0.21rc2)
ERROR: No matching distribution found for scikit-learn==0.22.2 (from adversarial-robustness-toolbox==1.4.2)
```

beat-buesser:
# Description

This pull request extends the test matrix to include recent versions of Python 3.8, scikit-learn 0.23.2 and TensorFlow 2.3.1 and 2.4.0rc1

Fixes #734 #587

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


.github/workflows/ci.yml
```diff
@@ -20,39 +20,59 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         include:
           - framework: tensorflow
```
ebubae, line 28: It may make sense to have a separate attribute here called `name` where we can give a short name to each test variant, so it makes sense when debugging. Ideally we wouldn't have two tests that report with the same name. To that end we could also probably drop the `tf_version` attribute

.github/workflows/ci.yml
```diff
@@ -20,39 +20,59 @@ on:
...
+            tensorflow: 2.2.1
             keras: 2.3.1
+            scikit-learn: 0.22.2
+          - framework: legacy
+            python: 3.8
+            tensorflow: 2.3.1
+            keras: 2.4.3
+            scikit-learn: 0.23.2
 
     name: Run ${{ matrix.framework }} ${{ matrix.tf_version }} Tests
```
ebubae, line 77: After adding short names to each variant, this line can be changed.
```suggestion
    name: Run ${{ matrix.name }} Tests
```

.github/workflows/ci.yml
```diff
@@ -63,11 +83,23 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Pre-install
```
ebubae, line 85: The names for each step should represent what is going on in this step. That way, if the step is skipped it makes sense.
```suggestion
      - name: Pre-install legacy
```

.github/workflows/ci.yml
```diff
@@ -63,11 +83,23 @@ jobs:
...
-        if: ${{ matrix.framework == 'tensorflow' || matrix.framework == 'keras' || matrix.framework == 'kerastf' || matrix.framework == 'legacy' }}
+        if: ${{ matrix.framework == 'legacy' }}
+        run: |
+          sudo apt-get -y -q install ffmpeg libavcodec-extra
+          pip install tensorflow==${{ matrix.tensorflow }}
+          pip install keras==${{ matrix.keras }}
+          pip install scikit-learn==${{ matrix.scikit-learn }}
+      - name: Pre-install
```
ebubae, line 92:
```suggestion
      - name: Pre-install tensorflow
```

.github/workflows/ci.yml
```diff
@@ -63,11 +83,23 @@ jobs:
...
+        if: ${{ matrix.framework == 'tensorflow' || matrix.framework == 'keras' || matrix.framework == 'kerastf' }}
+        run: |
+          sudo apt-get -y -q install ffmpeg libavcodec-extra
+          pip install tensorflow==${{ matrix.tensorflow }}
+          pip install keras==${{ matrix.keras }}
+      - name: Pre-install
```
ebubae, line 98:
```suggestion
      - name: Pre-install scikitlearn
```

ebubae (changes requested):
Thanks @beat-buesser for adding these changes to make our test cover more configurations. I only have a couple of suggestions.

ebubae:
Separate bug this PR also helped me realize: we need to change the push trigger to only run for `dev[version number]` and not just any branch that starts with `dev` like this one.

On line 15, could you replace `dev*` with `dev[0-9]+.[0-9]+.[0-9]` ?

.github/workflows/ci.yml
```diff
@@ -20,39 +20,59 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-20.04
```
ebubae, line 23: I kept getting inexplicable errors with MXNet tests (see explanation in #615) when I tried to use a later versions of Ubuntu. Just in case you end up running into issues with this like I did.

beat-buesser: Ok, I remember. Let's see if it works now. We also could see if we can upgrade `mxnet` as part of this PR to fix the problem.

beat-buesser:
> Separate bug this PR also helped me realize: we need to change the push trigger to only run for `dev[version number]` and not just any branch that starts with `dev` like this one.
> 
> On line 15, could you replace `dev*` with `dev[0-9]+.[0-9]+.[0-9]+` ?

I think we should leave as is, that we we can get test insights without having to open PRs if we start the branch name with dev*.

.github/workflows/ci.yml
beat-buesser: I think that makes sense, I'll add it in new commits.

ebubae:
@beat-buesser that's fine too if we want to set the convention.

I was suggesting this because once the pull request is eventually made (like this one) you get two sets of tests for `pull_request` and `push`. This is also the case for #736, but now we have many more tests.

One of the ways to get around this is by opening a draft PR. Any pushes on the PR will be re-run by Github Actions, and once the PR is ready it can be marked as ready for review or closed if we were just running tests.

If we want to have branches that are dedicated for testing but will not be merged, that is also fine.

beat-buesser:
@ebubae I see. I think for Travis the difference between `pull_request` and `push` is that the former was testing after merging the target branch whereas the latter was testing without merging which in some cases provided insights into bugs. I would like to propose to keep both for a while to see if they behave differently with Actions. I agree, we have now a larger number of runs and should keep them at a minimum but I'm also very happy that they run much faster.

ebubae (approved):
Thanks @beat-buesser for adding so many new builds to our test matrix. This will definitely help make sure future code works as expected with our supported frameworks.

PR #762, opened 2020-11-26
Edit clever_t to mitigate OOM and long runtime issues

shawnclq:
Signed-off-by: shawnclq <sclq1993@gmail.com>

# Description

## Summary
To mitigate long runtime issue, compute the class gradients of all samples in `rand_pool` before looping through the batches.
To avoid `ResourceExhasutedError`, batching is also done on `rand_pool` so that computation of class gradients is not done on the entire set of `rand_pool`, but instead on smaller batches of size `pool_factor`.

### `ResourceExhaustedError`
When I first started using this metric, I ran into `ResourceExhasutedError` when `batch_size` was too large (I set `batch_size = 100`), debugging the function uncovered the following lines as the cause.

`grad_pred_class = classifier.class_gradient(sample_xs, label=pred_class)`
`grad_target_class = classifier.class_gradient(sample_xs, label=target_class)`

A large `batch_size` would result in `sample_xs` being too large, and users might run into the same error if their computing resources are not sufficient.
**Initially**, I edited the `clever_t` function to introduce batching for `sample_xs`, as shown below:
```
sample_xs = rand_pool[np.random.choice(pool_factor * batch_size, batch_size)]
grad_norm = []
mini_batch_size = 5
# Compute gradients
for i in range(len(sample_xs) // mini_batch_size):
    sample_x = sample_xs[i * mini_batch_size:(i+1) * mini_batch_size]

    grads_pred_class = classifier.class_gradient(sample_x, label=pred_class)
    grads_target_class = classifier.class_gradient(sample_x, label=target_class)

    if np.isnan(grads_pred_class).any() or np.isnan(grads_target_class).any() :
        raise Exception("The classifier results NaN gradients.")

    grad = grads_pred_class - grads_target_class
    grad = np.reshape(grad, (mini_batch_size, -1))
    grad_norm.extend(np.linalg.norm(grad, ord=norm, axis=1))
grad_norm = np.max(np.array(grad_norm))
grad_norm_set.append(grad_norm)
```
This edit would avoid the `ResourceExhaustedError` to ensure that the size of `sample_x` is small enough to input into the function `class_gradient`.

### Long Runtime Issue
However, even with the above edit and with my computing resources (which are not good enough to begin with), the time taken to generate the CLEVER score for one image takes too long.
The main contributing factor to the long runtime are the class gradient computation lines as mentioned above.
I computed the time taken to generate the CLEVER score for one image and it took 980 seconds = 16 minutes. With 100 images, it would take 26 hours. (This amount of time taken varies greatly with the model complexity, size of image and computing resources).  

The goal is to minimise the number of times the function `class_gradient` is called.  
In the original code, the class gradient of the same sample might be calculated multiple times since it might be randomly chosen at different batches.
To avoid duplicate computations and to reduce the number of computations, the class gradient of all samples in `rand_pool` should be computed and saved in an array. Subsequently when looping through the batches, the values can then randomly selected from that array.
Batching is also done for `rand_pool`, with a batch size of `pool_factor` (since `pool_factor` is a factor of the size of `rand_pool`), so that the above `ResourceExhaustedError` can be avoided as well.   

This would greatly minimise the number of computations and reduce the time taken to generate the CLEVER score for an image, especially when `nb_batches` and `batch_size` are set to be large.
With this change, the time taken reduced from 980s to 57s, a 17x decrease.

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


beat-buesser:
Hi @shawnclq Thank you very much for your pull request and the detailed summary! We will review it as soon as possible in the coming days.

shawnclq:
Hi @beat-buesser, thanks! Do let me know if there’s anything else I can do to help make my code easier to review.

beat-buesser:
Hi @shawnclq Sorry for the delay, we were busy with the 1.5 release. I'm looking at now.

art/metrics/metrics.py
```diff
@@ -319,21 +320,29 @@ def clever_t(
     elif norm != 2:
         raise ValueError("Norm {} not supported".format(norm))
 
-    # Loop over the batches
-    for _ in range(nb_batches):
-        # Random generation of data points
-        sample_xs = rand_pool[np.random.choice(pool_factor * batch_size, batch_size)]
+    # Compute gradients for all samples in rand_pool
+    for i in range(batch_size):
+        rand_pool_batch = rand_pool[i * pool_factor:(i+1) * pool_factor]
```
beat-buesser, line 325: E226 missing whitespace around arithmetic operator
```suggestion
        rand_pool_batch = rand_pool[i * pool_factor:(i + 1) * pool_factor]
```

beat-buesser (changes requested):
Hi @shawnclq The changes look good to me. I was able to reproduce the speed up of the clever_t metric and I think your optimisation will be useful for all users of the Clever metric. I have one change request to pass the Python code Style Check, could you please take a look?

shawnclq:
Hi @beat-buesser , I have made the change, do take a look.
Thanks for taking your time to look through my code!

beat-buesser (approved):

PR #773, opened 2020-11-30
Add Lingvo test runner to CI

### LINKED ISSUE #707: Add test runner for Lingvo ASR estimator
The Lingvo ASR estimator requires a very specific combination of `lingvo` and `tensorflow-gpu` that is not supported on CI. Therefore, CI will skip those tests at the moment.

Given the work of @ebubae, who added Github Actions to our CI, we can now easily add new test runners. I would like to propose that we add a specific job for the Lingvo estimator.

hesseltuinhof:
# Description

This PR adds a Lingvo test runner to the CI pipeline. All tests related to Lingvo, require a specific version of `tensorflow` (`2.1.0`), Lingvo (`0.6.4`) and Python 3.6.

Note, the new test runner will run all `tensorflow2v1` flagged tests.

Fixes #707

## Type of change

**Test Configuration**:

- Fedora 33
- Python 3.6

# Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,13 @@ jobs:
           - name: mxnet (Python 3.7)
             framework: mxnet
             python: 3.7
+          - name: TensorFlow+Lingvo 2.1.0v1 (Keras 2.3.1 Python 3.6)
+            framework: tensorflow2v1
+            python: 3.6
+            tensorflow: 2.1.0
+            tf_version: v2
+            keras: 2.3.1
+            lingvo: 0.6.4
           - name: legacy (TensorFlow 2.2.1 Keras 2.3.1 scikit-learn 0.22.2 Python 3.7)
             framework: legacy
             python: 3.7
@@ -115,7 +122,7 @@ jobs:
           pip install scikit-learn==${{ matrix.scikit-learn }}
           pip list
       - name: Pre-install tensorflow
-        if: ${{ matrix.framework == 'tensorflow' || matrix.framework == 'keras' || matrix.framework == 'kerastf' }}
+        if: ${{ matrix.framework == 'tensorflow' || matrix.framework == 'keras' || matrix.framework == 'kerastf' || matrix.framework == 'tensorflow2v1' }}
         run: |
           pip install tensorflow==${{ matrix.tensorflow }}
           pip install keras==${{ matrix.keras }}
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -79,6 +79,14 @@ jobs:
           - name: mxnet (Python 3.7)
             framework: mxnet
             python: 3.7
+          - name: TensorFlow+Lingvo 2.1.0v1 (Keras 2.3.1 Python 3.6)
+            framework: tensorflow2v1
+            python: 3.6
+            tensorflow: 2.1.0
+            tenorflow_gpu: 2.1.0
```
beat-buesser, line 86: Is the label `tensorflow_gpu` still required? I think it's covered by label `tensorflow`.

setup.cfg
```diff
@@ -9,6 +9,7 @@ markers =
     only_with_platform: DEPRECATED only used for legacy tests. Use skipMlFramework instead. marks a test to be performed only for a specific mlFramework values
     framework_agnostic: marks a test to be agnostic to frameworks and run only for one default framework
     skip_travis: Skips a test marked with this decorator if the command line argument skip_travis is set to true
+    skipModule: Skip the test if a module is not available in the current environment
```
beat-buesser, line 12: How about adapting the marker name following the existing naming pattern? Although `skipMlFramework` is also an outlier.
```suggestion
    skip_module: Skip the test if a module is not available in the current environment
```

beat-buesser (changes requested):
Hi @hesseltuinhof Thank you very much for your pull request! I like the changes and have a few question which might add additional improvements. What do you think?

.github/workflows/ci.yml
hesseltuinhof: Yes, you are right. 

I also checked the `setup.py` of the Lingvo library and they install `tensorflow-gpu` via `tensorflow-gpu==tf.__version__` (see [setup.py#L43](https://github.com/tensorflow/lingvo/blob/55263937599fdbbfb959e8eab7a40fa0aa74ad9c/pip_package/setup.py#L43)). So we can stick to the same snippets we use for the other Tensorflow workflows.

setup.cfg
hesseltuinhof: The original code to skip modules uses camel case. I agree that both `skipModule` and `skipMlFramework` are outliers. So let's start to fix the `skip_module` with this PR :relaxed:

I will open an issue for `skipMlFramework` and if time allows update it as well :+1:

hesseltuinhof:
Thanks for suggesting the two changes. I fully agree with them and have applied the PR accordingly :+1:

beat-buesser (approved):
@hesseltuinhof Thank you!

PR #781, opened 2020-12-02
Implement set_learning_phase as no-op

### LINKED ISSUE #779: Implement set_learning_phase method(s) as no-op where applicable
When running training and then inference on an estimator, some estimators need `set_learning_phase` set at training and again at inference, while others do not. If `set_learning_phase` is not called, some estimators will not function properly, and for other estimators, if `set_learning_phase` is called, it will raise a NotImplementedError.

Instead of raising NotImplementedError, I would suggest it performing a no-op for models where set_learning_phase is irrelevant (i.e., for those models where fit and predict don't require any change), and it only raising an error if you try to set the learning phase to something not supported (e.g., `fixed_model.set_learning_phase(True)`).

To handle an arbitrary model for a particular task then requires introspection on the type of model as to whether to call set_learning_phase or not. However, this breaks some of the estimator abstractions and can add a lot of case logic to a simple script. 

Another way to handle this would be some boolean logic to query whether the model can handle a particular phase, like `is_learning_phase` or `is_inference_phase`, which can then be queried, and then `set_learning_phase` can be applied afterward. However, this would make even more functions and complexity.

**Additional context**
Context is using armory scenarios with training and inference - we use set_learning_phase when training is used, which causes errors for TensorFlowV2Classifier and potentially others.

beat-buesser:

# Description

This pull request implements `set_learning_phase` as no-op where applicable. 

Fixes #779 

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/art/estimators/classification/tensorflow.py
+++ b/art/estimators/classification/tensorflow.py
@@ -1146,11 +1146,15 @@ class TensorFlowV2Classifier(ClassGradientsMixin, ClassifierMixin, TensorFlowV2E
 
     def set_learning_phase(self, train: bool) -> None:
         """
+        Calling this method does not result in any operations or changes and simply passes. This method is implemented
+        to enable calling `set_learning_phase` independent of the estimator type and framework-specific requirements to
+        set the learning phase. For TensorFlow v2 the learning phase has not be defined.
+
         Set the learning phase for the backend framework.
 
         :param train: True to set the learning phase to training, False to set it to prediction.
         """
-        raise NotImplementedError
+        pass
 
     def save(self, filename: str, path: Optional[str] = None) -> None:
         """
--- a/art/estimators/encoding/tensorflow.py
+++ b/art/estimators/encoding/tensorflow.py
@@ -149,7 +149,8 @@ class TensorFlowEncoder(EncoderMixin, TensorFlowEstimator):  # lgtm [py/missing-
         """
         Do nothing.
         """
-        raise NotImplementedError
+        if train:
+            raise NotImplementedError("Training is not implemented for this estimator.")
 
     def loss_gradient(self, x: "np.ndarray", y: "np.ndarray", **kwargs) -> "np.ndarray":
         """
--- a/art/estimators/generation/tensorflow.py
+++ b/art/estimators/generation/tensorflow.py
@@ -160,9 +160,10 @@ class TensorFlowGenerator(GeneratorMixin, TensorFlowEstimator):  # lgtm [py/miss
 
     def set_learning_phase(self, train: bool) -> None:
         """
-        do nothing.
+        Do nothing.
         """
-        raise NotImplementedError
+        if train:
+            raise NotImplementedError("Training is not implemented for this estimator.")
 
     @property
     def model(self) -> "tf.Tensor":
--- a/art/estimators/object_detection/pytorch_faster_rcnn.py
+++ b/art/estimators/object_detection/pytorch_faster_rcnn.py
@@ -325,7 +325,11 @@ class PyTorchFasterRCNN(ObjectDetectorMixin, PyTorchEstimator):
         raise NotImplementedError
 
     def set_learning_phase(self, train: bool) -> None:
-        raise NotImplementedError
+        """
+        Do nothing.
+        """
+        if train:
+            raise NotImplementedError("Training is not implemented for this estimator.")
 
     def loss(self, x: np.ndarray, y: np.ndarray, **kwargs) -> np.ndarray:
         """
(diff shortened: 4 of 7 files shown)


davidslater (changes requested):
For pytorch, since `model.train(train: bool)` is ubiquitous across models, I think it makes more sense to migrate `set_learning_phase` from `PyTorchClassifier` to `PyTorchEstimator`, as the former inherits from the latter. This would also make it easier to add new PyTorch models like a trainable ASR model without rewriting that method for each. Similarly, if users create their own PyTorch estimators and wrap them with PyTorchEstimator, they would need to reimplement that.

For similar reasons, I would recommend moving the TensorFlow functions to the estimator class. (I understand that you may want to override that for the classifier class, due to the use of the feed_dict for V1, but I think that will make it easier to subclass from estimator.)

I think a similar approach could apply to mxnet, keras, and scikitlearn estimators.

davidslater:
Let me know when this is up for another review.

beat-buesser:
Hi @davidslater I think this PR is ready for another review.

art/estimators/classification/tensorflow.py
```diff
@@ -800,7 +815,7 @@ def predict(self, x: np.ndarray, batch_size: int = 128, **kwargs) -> np.ndarray:
             )
 
             # Run prediction
-            results[begin:end] = self._model(x_preprocessed[begin:end])
```
davidslater, line 803: This and the following similar changes look like a bug to me, as it's now referencing `self.model` instead of `self._model`. Is this intentional?

davidslater (changes requested):
I have a few comments and thoughts:
1) Maybe we should move this to dev_1.6.0 instead of dev_1.5.1, since this does break backwards compatibility with the `set_learning_phase` function. External code that currently calls it will fail. An alternative that would make things break less would be to add `set_learning_phase` as a no-op in the base class art.estimators.BaseEstimator or NeuralNetworkMixin that has a deprecation warning associated with it. (You can say that it will be removed in 1.6.0, for instance.) Something like this:
```
    @deprecated(end_version="1.6.0", replaced_by="kwarg 'training_mode' in 'class_gradient' and 'loss_gradient' methods")
    def set_learning_phase(self, train: bool) -> None:                                           
        pass
```
2) For wrapper-type classes, like detection/evasion/detector.BinaryActivationDetector and classification ensemble, internal calls to class_gradient and loss_gradient need to pass `**kwargs` through so that if the underlying classifier takes the training_mode kwarg, it gets properly passed through.
3) We may want to consider adding the "training_mode" parameter to predict, as that would be consistent with the way TF2 is now operating. Plus, you may want to be able to easily compare predictions with and without something like batchnorm or dropout. Similarly on expectation.py and potentially others.
4) I tested with keras and pytorch, but I have never used mxnet, and don't have tf graph or tf eager models to test with.
5) Is there a rationale for placing a lot of the functionality in the classifier classes instead of the estimator base classes?

art/estimators/classification/tensorflow.py
beat-buesser: I noticed that we were not using the new model property everywhere and made the updates, but now I think it's better to make these changes in another pull request and I have reverted them here.

art/estimators/classification/keras.py
```diff
@@ -373,15 +375,24 @@ def loss(self, x: np.ndarray, y: np.ndarray, reduction: str = "none", **kwargs)
...
 
         :param x: Sample input with shape as expected by the model.
         :param y: Target values (class labels) one-hot-encoded of shape (nb_samples, nb_classes) or indices of shape
                   (nb_samples,).
+        :param training_mode: `True` for model set to training mode and `'False` for model set to evaluation mode.
         :return: Array of gradients of the same shape as `x`.
         """
+        # pylint: disable=E0401
+        if self.is_tensorflow:
+            import tensorflow.keras.backend as k
```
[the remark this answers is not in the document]
beat-buesser: Yes, I think it's the right time for this update for learning phase.

`KerasClassifier` contains a lot of code that is necessary for earlier version of `keras` whereas since Keras 2.3 we recommend using `TensorFlowV2Classifier` for Keras models using `tf.keras` where we are not relying on `set_learning_phase`. However the name of `TensorFlowV2Classifier` can be confusing for users of Keras models. I would propose that if `KerasClassifier` is used for `keras>=2.3` that we internally replace it with `TensorFlowV2Classifier`.

beat-buesser:
Hi @davidslater Thank you very much for your great and useful review. I completely agree with your proposals and I'll add updates accordingly.

About item 5.: The only rational would be that the classifier classes were part of ART from the beginning well before the estimator abstraction was introduced. New shared and reusable code has to go to the highest abstraction in the Estimator API, but it might still be the case that code is duplicated on the classification level and we aim to add updates where possible. Are you thinking of a specific part?

PR #791, opened 2020-12-08
Test codecov with GitHub Actions

beat-buesser:
# Description

This pull request tests codecov with GitHub Actions.


ebubae (approved):
These current changes allow us to integrate codecov with the new ART testing framework using pytest. Per-framework coverage reports are now saved at the end of each run.

PR #796, opened 2020-12-11
#580 evasion tests integration_C

### LINKED ISSUE #580: Move all Evasion tests to the new ART Testing Framework and make them framework independent
Most evasion tests are still running as framework specific tests, which contributes to reaching the Travis test limit.

Move all Evasion tests to the new ART Testing Framework and make them framework independent, this will significantly reduce the time required to run all the tests on Travis

killianlevacher:
# Description

This PR addresses issue #580 which is about integrating remaining evasion tests within the new pytest framework

Fixes #580 evasion

## Type of change

- [X] Improvement (non-breaking)
- [ ] Bug fix (non-breaking)

# Checklist


conftest.py
```diff
@@ -138,7 +138,7 @@ def _image_dl_estimator_for_attack(attack, defended=False, **kwargs):
             raise ARTTestFixtureNotImplemented(
                 "no estimator available", image_dl_estimator_for_attack.__name__, framework, {"attack": attack}
             )
-        return classifier_tested[0]
+        return classifier_tested[0], sess
```
beat-buesser, line 141: Why is this change required? The returned `sess` does not seem to be used anywhere.

beat-buesser (changes requested):

conftest.py
killianlevacher: ye it's rarely used indeed but the session is needed some times for Tensorflow models. That's a discrepancy that I hadn't seen before. The session is used in some test cases when `image_dl_estimator` is called directly, hence `image_dl_estimator_for_attack` which is just an augmented version of  `image_dl_estimator` by convention should have passed on that session too from the start. That was something I had overlooked so I just felt I should fix this now rather than later when someone complains why isn't it passed on :-)

PR #883, opened 2021-02-03
Check for NaN in loss gradient of PGD attacks and replace with 0.0

### LINKED ISSUE #824: Better PGD handling of NaN and Inf gradient outputs
Currently, if the gradient produces a single nan during PGD (at least for `norm != inf`), it propagates to the rest of the produced adversarial sample (and in PyTorch, it crashes). This happens most clearly here:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py#L284-L293
While `.sign()` in torch directly maps nan to 0, the normalization for L1 and L2 will make `grad` equal to nan for all values if a single element is nan.

A) Looks for NANs and generate warnings when they occur.
B) Set NANs to 0 when they do occur.
In torch, a simple solution to B that solves my specific problem would be to add this before L284:
```
grad[grad.isnan()] = 0
```
There are some edge cases where infinite values and masks could also generate nans, and I can walk through some of those as well.

The primary alternative would be to modify the estimator. While most model gradients and typically well behaved with respect to the weights, they are not always so well behaved with respect to the input, and it can be difficult to remove nan-generating edge cases from models. As an example, in one recent model I was working with, if the audio sample was zero-padded, it would generate nan gradients for the first section of the input. However, I do not have an easy way to modify the model to remove that issue, so this would be a difficult process.

beat-buesser:

# Description

This pull request check for elements with NaN in the loss gradient of all PGD attacks and replaces them with 0.0. A warning is logged if an element with NaN is detected.

Fixes #824

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/art/attacks/evasion/fast_gradient.py
+++ b/art/attacks/evasion/fast_gradient.py
@@ -344,7 +344,7 @@ class FastGradientMethod(EvasionAttack):
         # Check for NaN before normalisation an replace with 0
         if np.isnan(grad).any():
             logger.warning("Elements of the loss gradient are NaN and have been replaced with 0.0.")
-            grad = np.where(np.isnan(grad), np.zeros_like(grad), grad)
+            grad = np.where(np.isnan(grad), 0.0, grad)
 
         # Apply mask
         if mask is not None:
@@ -352,6 +352,9 @@ class FastGradientMethod(EvasionAttack):
 
         # Apply norm bound
         def _apply_norm(grad, object_type=False):
+            if np.isinf(grad).any():
+                logger.info("The loss gradient array contains at least one positive or negative infinity.")
+
             if self.norm in [np.inf, "inf"]:
                 grad = np.sign(grad)
             elif self.norm == 1:
--- a/art/attacks/evasion/fast_gradient.py
+++ b/art/attacks/evasion/fast_gradient.py
@@ -341,6 +341,11 @@ class FastGradientMethod(EvasionAttack):
         # Get gradient wrt loss; invert it if attack is targeted
         grad = self.estimator.loss_gradient(batch, batch_labels) * (1 - 2 * int(self.targeted))
 
+        # Check for NaN before normalisation an replace with 0
+        if np.is_nan(grad).any():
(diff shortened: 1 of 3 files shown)


art/attacks/evasion/fast_gradient.py
```diff
@@ -341,6 +341,11 @@ def _compute_perturbation(
         # Get gradient wrt loss; invert it if attack is targeted
         grad = self.estimator.loss_gradient(batch, batch_labels) * (1 - 2 * int(self.targeted))
 
+        # Check for NaN before normalisation an replace with 0
+        if np.isnan(grad).any():
+            logger.warning("Elements of the loss gradient are NaN and have been replaced with 0.0.")
+            grad = np.where(np.isnan(grad), np.zeros_like(grad), grad)
```
davidslater, line 347: `np.zeros_like(grad)` can just be 0.0 (it will be broadcast by the method).

davidslater (changes requested):
The main issue I see is in `_apply_norm`. If the perturbation has any `inf` or `-inf` values in it, the resultant vector will be all 0s and NaNs:
```
>>> grad
array([-inf, -10.,  -1.,   0.,   1.,  10.,  inf,   0.])
>>> grad / (np.sum(np.abs(grad), axis=ind, keepdims=True) + tol)
array([nan, -0., -0.,  0.,  0.,  0., nan,  0.])
>>> grad / (np.sqrt(np.sum(np.square(grad), axis=ind, keepdims=True)) + tol)
array([nan, -0., -0.,  0.,  0.,  0., nan,  0.])
```

Fixing this problem is difficult, as you need to define `inf / inf`, but at the least you should have a warning if infinite values appear in the gradient.

art/attacks/evasion/fast_gradient.py
beat-buesser: Good idea.

davidslater (approved):
LGTM

PR #891, opened 2021-02-10
Implement method to apply adversarial patch onto a real-world surface

### LINKED ISSUE #853: Extend mask feature in AdversarialPatch*.apply_patch to include perspective transformation
We would like to extend the method `apply_patch` of the Adversarial Patch attacks to place the patch on a predefined area in the image, similar to `mask`, but also to recognise and apply the perspective transformations of the patch required to maximise overlap. It should also be possible to apply similar but randomly selected transformations during training.

beat-buesser:
# Description

This pull request adds a method to apply the patch to coordinates after perspective transformation. These coordinates can be the coordinates of a real-world, quadratic surface in the targeted image.

Thank you to @yusong-tan for sharing the code!

Fixes #853 

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/art/attacks/evasion/adversarial_patch/adversarial_patch.py
+++ b/art/attacks/evasion/adversarial_patch/adversarial_patch.py
@@ -31,9 +31,9 @@ import numpy as np
 
 from art.attacks.evasion.adversarial_patch.adversarial_patch_numpy import AdversarialPatchNumpy
 from art.attacks.evasion.adversarial_patch.adversarial_patch_tensorflow import AdversarialPatchTensorFlowV2
+from art.attacks.evasion.adversarial_patch.utils import insert_transformed_patch
 from art.estimators.estimator import BaseEstimator, NeuralNetworkMixin
 from art.estimators.classification.classifier import ClassifierMixin
-
 from art.estimators.classification import TensorFlowV2Classifier
 from art.attacks.attack import EvasionAttack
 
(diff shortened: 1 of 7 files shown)


art/attacks/evasion/adversarial_patch/adversarial_patch.py
```diff
@@ -169,6 +168,19 @@ def reset_patch(self, initial_patch_value: Optional[Union[float, np.ndarray]]) -
         """
         self._attack.reset_patch(initial_patch_value=initial_patch_value)
 
+    def insert_transformed_patch(self, x: np.ndarray, patch: np.ndarray, image_coords: np.ndarray):
+        """
+        Insert patch to image based on given or selected coordinates.
+
+        :param x: The images to insert the patch.
```
minhitbk, line 175: I think this function is only applied for a single image, so this description confuses me that the patch can be applied for a batch of images.

art/attacks/evasion/adversarial_patch/adversarial_patch_numpy.py
```diff
@@ -578,3 +579,17 @@ def reset_patch(self, initial_patch_value: Optional[Union[float, np.ndarray]]) -
             self.patch = initial_patch_value
         else:
             raise ValueError("Unexpected value for initial_patch_value.")
+
+    @staticmethod
+    def insert_transformed_patch(x: np.ndarray, patch: np.ndarray, image_coords: np.ndarray):
+        """
+        Insert patch to image based on given or selected coordinates.
+
+        :param x: The images to insert the patch.
```
minhitbk, line 588: I think this function is only applied for a single image, so this description confuses me that the patch can be applied for a batch of images.

art/attacks/evasion/adversarial_patch/adversarial_patch_tensorflow.py
```diff
@@ -495,3 +496,17 @@ def reset_patch(self, initial_patch_value: Optional[Union[float, np.ndarray]] =
             self._patch.assign(initial_patch_value)
         else:
             raise ValueError("Unexpected value for initial_patch_value.")
+
+    @staticmethod
+    def insert_transformed_patch(x: np.ndarray, patch: np.ndarray, image_coords: np.ndarray):
+        """
+        Insert patch to image based on given or selected coordinates.
+
+        :param x: The images to insert the patch.
```
minhitbk, line 505: I think this function is only applied for a single image, so this description confuses me that the patch can be applied for a batch of images.

minhitbk (approved):
Hi @beat-buesser, your code looks fine. I only have one small comment and one suggestion: that is the function `insert_transform_patched` in the end is refered to just one utility function, so it is not necessary to create an intermediate call in the common class and then the 2 other classes call the same function. We can do directly in the common class and therefore simplify the code. However, I respect your decision with your design if you think it is needed to keep it this way.

beat-buesser:
Hi @minhitbk Thank you very much for your review! I have updated the docstrings. The current approach with the static methods aims to enable users to directly use the specific implementations.

PR #901, opened 2021-02-16
Update set_params in BaseEstimator

### LINKED ISSUE #888: BaseEstimator.set_params not updating self.preprocessing with preprocessing_defences in ART 1.5
**Describe the bug**
ART 1.5 is internally combining `preprocessing` and `preprocessing_defences` into self.preprocessing. The method `BaseEstimator.set_params` has not been updated to account for this change. This results in preprocessing defences set with `set_params` are not being recognised and not being applied by all estimators.

**Expected behavior**
Treat preprocessing defences set with `set_params` to be treated equally as when defined in `__init__`.

This issue affects only ART 1.5, earlier version should not be affected.

beat-buesser:

# Description

This pull request updates `BaseEstimator.set_params` to follow changes in ART 1.5 to `preprocessing` and `preprocessing_defences` and set them correctly.

Fixes #888

## Type of change

- [x] Bug fix (non-breaking)

# Checklist

### DIFF
--- a/art/estimators/estimator.py
+++ b/art/estimators/estimator.py
@@ -47,7 +47,7 @@ class BaseEstimator(ABC):
         "clip_values",
         "preprocessing_defences",
         "postprocessing_defences",
-        # "preprocessing", # preprocessing cannot be set with set_params
+        "preprocessing",
     ]
 
     def __init__(
@@ -56,7 +56,7 @@ class BaseEstimator(ABC):
         clip_values: Optional["CLIP_VALUES_TYPE"],
         preprocessing_defences: Union["Preprocessor", List["Preprocessor"], None] = None,
         postprocessing_defences: Union["Postprocessor", List["Postprocessor"], None] = None,
-        preprocessing: "PREPROCESSING_TYPE" = (0, 1),
+        preprocessing: Union["PREPROCESSING_TYPE", "Preprocessor"] = (0, 1),
     ):
         """
         Initialize a `BaseEstimator` object.
@@ -77,7 +77,7 @@ class BaseEstimator(ABC):
         self._model = model
         self._clip_values = clip_values
 
-        self.preprocessing = preprocessing
+        self.preprocessing = self._set_preprocessing(preprocessing)
         self.preprocessing_defences = self._set_preprocessing_defences(preprocessing_defences)
         self.postprocessing_defences = self._set_postprocessing_defences(postprocessing_defences)
         self.preprocessing_operations: List["Preprocessor"] = []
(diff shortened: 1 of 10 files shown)


art/estimators/estimator.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         clip_values: Optional["CLIP_VALUES_TYPE"],
         preprocessing_defences: Union["Preprocessor", List["Preprocessor"], None] = None,
         postprocessing_defences: Union["Postprocessor", List["Postprocessor"], None] = None,
-        preprocessing: "PREPROCESSING_TYPE" = (0, 1),
+        preprocessing: Union["PREPROCESSING_TYPE", "Preprocessor"] = (0, 1),
```
minhitbk, line 59: If I understand correctly, the type of "Preprocessor" is applied for preprocessing_defences while the type "PREPROCESSING_TYPE" is applied for the preprocessing param for standardizing the input. So why now you add this type for preprocessing param, is there any design change?

art/estimators/estimator.py
```diff
@@ -246,7 +294,7 @@ def _apply_preprocessing(self, x, y, fit: bool) -> Tuple[Any, Any]:
         :rtype: Format as expected by the `model`
         """
         if self.preprocessing:
```
minhitbk, line 296: Should this be `if self.preprocessing_operations`?

minhitbk (approved):
I think the changes are fine. There are only a minor comment. In my opinion, the current way we control preprocessing and preprocessing_defences is complex. We now already have StandadizationPreprocessor, so we should stop supporting preprocessing as tuple (mean, std) and then we can remove a lot of complexity, making the code cleaner and easy understand (I mean in future release), for example we only need 1 att preprocessing_defence instead of 3 (preprocessing, preprocessing_defences and preprocessing_operations).

art/estimators/estimator.py
beat-buesser: Yes, that's correct! Thank you.

art/estimators/estimator.py
beat-buesser: Yes, that's correct. We are now internally converting the preprocessing tuple to an instance of Preprocessor. One reason is to enable full framework support through the preprocessing steps. Because of the conversion `get_params` is returning the Preprocessor instance instead of the tuple and therefore the argument `preprocessing` and especially `set_params` accept now also the Preprocessor type for `preprocessing`. The tuple remains for backward compatibility.

beat-buesser:
Hi @minhitbk Thank you very much for the great review. I agree on the increased complexity and your proposals to simplify. As you mentioned this added complexity is to guarantee backward compatibility.

PR #912, opened 2021-02-22
Fix for changing DeepSpeech states

### LINKED ISSUE #908: DeepSpeech state changes after multiple loss_gradient calls
The DeepSpeech estimator returns different predictions after several `loss_gradient` calls. It seems that somehow the state of the model changes and does not persist.

The following  pattern changes the  initial  state of the DeepSpeechestimator:

* Step 1: Load audio example and run a prediction
* Step 2: Multiple calls to loss gradient method w.r.t.random input
* Step 3: Re-run prediction on example of Step 1
* Result: Different predictions  for Step 3

The expected result should be that both steps 1 and 3 return the exact same prediction.

 - Fedora 33
 - Python version 3.6.10
 - ART 1.5.1
 - PyTorch 1.5.1 with DeepSpeech 2.1

hesseltuinhof:
# Description

This fixes the changing DeepSpeech states that can occur after multiple `loss_gradient` calls.

Fixes #908 

## Type of change

- [x] Bug fix (non-breaking)

# Testing

I tested this against the script that was used to reproduce the problem, which was described in the linked issue.

**Test Configuration**:
- Fedora 33
- Python 3.6.10
- ART 1.5.1
- PyTorch 1.5.1 with DeepSpeech 2.1

# Checklist

### DIFF
--- a/art/attacks/evasion/imperceptible_asr/imperceptible_asr_pytorch.py
+++ b/art/attacks/evasion/imperceptible_asr/imperceptible_asr_pytorch.py
@@ -25,18 +25,17 @@ specifically for PyTorch.
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import logging
-
-from typing import Tuple, Optional, TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional, Tuple
 
 import numpy as np
 import scipy
 
 from art.attacks.attack import EvasionAttack
+from art.config import ART_NUMPY_DTYPE
 from art.estimators.estimator import BaseEstimator, LossGradientsMixin, NeuralNetworkMixin
 from art.estimators.pytorch import PyTorchEstimator
-from art.estimators.speech_recognition.speech_recognizer import SpeechRecognizerMixin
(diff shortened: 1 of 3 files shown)


beat-buesser (changes requested):
Hi @hesseltuinhof Thank you very much for discovering and providing a solution for the influence of unfrozen batch norm and dropout layers in some PyTorch models!

The code changes look good to me.

Could other attacks also benefit from these new method in `PyTorchEstimator`? I'm thinking for example of `ProjectedGradientDescentPyTorch`.

Would it make sense to add calls to `set_batchnorm` and `set_dropout` in `PyTorchEstimator.__init__` and let the user deactivate them with a new argument? Could this fix the issue for all attacks?

hesseltuinhof:
Thanks for your review. I will have a look into #781 as well. Probably, we want to catch possible PyTorch RNN models in that PR as well. We could then use the methods from this PR to freeze batchnorm and Dropout layers manually as we won't be able to put the whole model in eval mode! Thanks for the pointer!

beat-buesser (approved):
Hi @hesseltuinhof Thank you very much for noticing and fixing the issue of unfrozen layers in the ASR estimators!

PR #990, opened 2021-03-24
Fixes Bug in Pixel Attack and Threshold Attack. 

### LINKED ISSUE #960: Bug in Pixel Attack and Threshold Attack.
When th=None (to find minimal perturbation), the generate function returns only adversarial images for th=1 and not for any other thresholds. 

The bug arises due to [line number 165](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/7c6326905aed874eed23e6c35c2965396bdf25ab/art/attacks/evasion/pixel_threshold.py#L165) in pixel_threshold.py file which forgets the 2nd best adversarial sample. 

Current Line 164-173

```
while True:
                    image_result: Union[List[np.ndarray], np.ndarray] = []      # BUG IN THIS LINE !!!
                    threshold = (start + end) // 2
                    success, trial_image_result = self._attack(image, target_class, threshold, max_iter)
                    if image_result or success:
                        image_result = trial_image_result
                    if success:
                        end = threshold - 1
                    else:
                        start = threshold + 1
```

Possible solution to the bug:

```
image_result: Union[List[np.ndarray], np.ndarray] = []
while True:
                    threshold = (start + end) // 2
                    success, trial_image_result = self._attack(image, target_class, threshold, max_iter)
                    if success:
                        image_result = trial_image_result
                        end = threshold - 1
                    else:
                        start = threshold + 1
```

shashankkotyan:
Fixes #960
Fixes the handling of preprocessed images to convert them to raw images before attacking.
Adds max_iter parameter to the intialiser. 
Corrects the default parameters for Pixel Attack and Threshold Attack.


art/attacks/evasion/pixel_threshold.py
```diff
@@ -148,35 +172,30 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int
...
-        if np.max(x) <= 1:
-            scale_input = True
-        else:
-            scale_input = False
+        
+        scale_input = False if np.max(x) > 1 else True
 
         if scale_input:
-            x = x * 255.0
+            x = self.depreprocess_images(x)
```
beat-buesser, line 179: This step of de-preprocessing should not be required in the attack's code. The input provided to `generate` is already expected to be un-preprocessed. Scaling with the `clip_values` (e.g. 255) as before is ok if the attack code only works for a different pixel range.

beat-buesser: If preprocessing is required for the model this preprocessing should be provided directly to the estimator under attack.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -185,7 +204,7 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int
         adv_x_best = np.array(adv_x_best)
 
         if scale_input:
-            adv_x_best = adv_x_best / 255.0
+            adv_x_best = self.estimator._apply_preprocessing(adv_x_best, y)
```
beat-buesser, line 207: This should not be necessary, similar comment as above.

beat-buesser (changes requested):
Hi @shashankkotyan Thank you very much for you updates for Pixel and Threshold attacks. I have one comment about the new de-preprocessing which should not be necessary. I would like to see if we can find an alternative solution, what do you think?

shashankkotyan:
Hi @beat-buesser! 
Sure! I am open to any other possibility for streamlining the implementation. 

As per my experimentation and analysis, I noticed that DE and CMA-ES work better if the given input image has pixels in the integer domain i.e., in Range (0,255). 
If the input image has pixels in the float domain i.e., in Range (0,1), the performance of the evolutionary algorithm decreases substantially. (Due to obvious reasons) 

Therefore, previously I manually converted back the input to the integer domain and then reconverted the adversarial image in the float domain in case the given input image is in the float domain. 
However, I noticed the bug that neural networks are sensitive to preprocessing used to train so previously I updated the code to include the estimators preprocessing. 

Now, when you point at it, I guess, if the user is giving images in the float range the preprocessing of the estimator will be (0,1) and this de-processing of images wouldn't make sense. 

I removed the de-processing and reprocessing of images, instead, I added a logger warning if someone tries to use the attacks for images in the float domain.

beat-buesser:
Hi @shashankkotyan 

Thank you very much for the updates. I think we probably could even leave the scaling to [0, 255] range and back (the original lines 149-158 and 187-189) as we have it right now if the input is in [0, 1] range. These are the two common input ranges in ART.

If a model requires normalisation to a different range it will be handled by the estimator and its mean and std preprocessing.

What do you think about adding the scaling back and add raising a value error if the range of `x` is not [0, 1] or [0, 255] instead?

shashankkotyan:
> I think we probably could even leave the scaling to [0, 255] range and back (the original lines 149-158 and 187-189) as we have it right now if the input is in [0, 1] range. These are the two common input ranges in ART.
> 
> If a model requires normalisation to a different range it will be handled by the estimator and its mean and std preprocessing.
> What do you think about adding the scaling back and add raising a value error if the range of x is not [0, 1] or [0, 255] instead?

I am not sure about this, since if the input is already preprocessed then I don't think estimator's preprocessing will have values associated. It will be probably placeholder values as (0,1) to not reprocess the preprocessed image. In this case, if we had the previous scaling then the estimator will give most probably misclassify and this will portray that the adversarial attack was successful. 

I think a warning that the method is untested for preprocessed images is well enough.

beat-buesser:
Hi @shashankkotyan 

Ok, I see. Scaling the input is currently not possible, because you would be using the `estimator.predict` method on the scaled input. Have you considered scaling the attack paramters like the 127 in https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/ac8022beb0791bef19660a53a7266eae69e6705e/art/attacks/evasion/pixel_threshold.py#L164 or the `-1` in https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/ac8022beb0791bef19660a53a7266eae69e6705e/art/attacks/evasion/pixel_threshold.py#L173 to the input range?

shashankkotyan:
As per my experimentation and analysis, I noticed that DE and CMA-ES work better if the given input image has pixels in the integer domain i.e., in Range (0,255).
If the input image has pixels in the float domain i.e., in Range (0,1), the performance of the evolutionary algorithm decreases substantially. 

So even if we scale the threshold limits (for the Threshold Attack), I don't see any benefit due to the above-mentioned problem. 
Moreover, scaling of threshold would only benefit Threshold Attack where `th` variable behaves the same as epsilon for other LInfinity Attacks.
But, for Pixel Attack `th` variable is in the integer range since for Pixel Attack `th` variable sets the number of pixels to be perturbed.

beat-buesser:
Hi @shashankkotyan 
Thank you very much for the explanation, in that case I would recommend to replace the `logger.warning` in current line 159 with `raise ValueError(...)` because the attack seems to now expect range [0, 255] as input.

beat-buesser:
Hi @shashankkotyan 

The ValueError update looks good. I think the tests for Pixel and Threshold Attack need now to be updated accordingly. You can fix the python style check with Black formatter (https://black.readthedocs.io/en/stable/) using `black --line-length 120 art/`.

beat-buesser:
Hi @shashankkotyan Thank you very much for the updates. Could you please check you last commit and maybe revert it? It modifies the format of a large number of unrelated files. Could you please try `black==20.8b1` with command `lack --line-length 120 art/attacks/evasion/pixel_threshold.py`.

shashankkotyan:
@beat-buesser Any updates here? I am seeing some unsuccessful checks here...

beat-buesser:
Hi @shashankkotyan

It looks like there are some style check warnings, mainly too long lines. Could you please take a look?

Please apologise the delay. I understand that the implementation of the optimisation is written for `x` in range [0, 255]. I'm still thinking how to make it possible to support inputs `x` in range [0, 1] and [0, 255]. I'll try to make some code suggestions soon.

shashankkotyan:
Hi @beat-buesser!

Thanks for letting me know your concerns. 

Regarding the processing for input in the range[0,1]. 

Let me formalise the cases, 

#### Case 1: Inputs in range[0,255] (Case where ART library handles the preprocessing)
In this case, as the user gave inputs in the range[0,255], we can assume, their estimator also expects the input in the same range. As the attack itself is efficient in the input range, it works!

#### Case 2: Inputs not in the range[0,255]  (Case where ART library does not handle the preprocessing)
In this case, as the user gave inputs in the range other than [0,255], we can assume that the user does the preprocessing of the image outside of the ART library. Hence, we can assume, their estimator also expects the input in the same range. 
But, the attack parameters themselves are not well defined for the input range other than [0,255].

In this case, we can first manually rescale the input in the range[0,255] to make it work with the DE and CMA-ES. 
Then, before calling `estimator.predict()`, we can rescale the input from the range[0,255] to the original scale and then finally rescaling the `adv_x` variable also the original scale before returning.

beat-buesser:
Hi @shashankkotyan Please apologise my delayed response, I forgot to respond immediately. I like your proposal in you last post very much. I think the input to the three calls to method  `estimator.predict` are the main place where we have to ensure the original range and undo the scaling if applied. Could you please update the branch accordingly?

beat-buesser:
Hi @shashankkotyan I have changed the target branch to `dev_1.7.0` after ART 1.6.2 has been released. How is going with the discussed updates? Let me know if I can help!

shashankkotyan:
Hi @beat-buesser! Apologies for the no updates lately. For the past two weeks, I have been busy with semester assignments and submissions. 
I will push the updates to the fork somewhere in the next week. 

I sincerely apologise for the delay!

beat-buesser:
Hi @shashankkotyan Thank you very much for the immediate response and no worries, definitely focus first on your semester assignments and submissions! Let me know when I can help with commits or with something else.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -150,13 +151,21 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> n
...
+                This takes substainally long time to process. \
+                For sanity check, pass th=10 to the Attack instance."
             )
 
+        # NOTE: Pixel and Threshold Attacks are well defined for unprocessed images where the pixel values are,
+        #       8-Bit color i.e., the pixel values are np.uint8 in range [0, 255].
+
+        # TO-DO: Better checking of input image.
+        #        All other cases not tested needs the images to be rescaled to [0, 255].
         if np.max(x) <= 1:
```
shashankkotyan, line 164: Can this checking be better? 

I thought to use dtype of x to check but the input can also be in float as in [0.0, 255.0]. So not sure about it.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -18,7 +18,6 @@
 """
 This module implements the Threshold Attack and Pixel Attack.
 The Pixel Attack is a generalisation of One Pixel Attack.
-
```
beat-buesser, line 21: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -60,42 +60,46 @@
 class PixelThreshold(EvasionAttack):
     """
     These attacks were originally implemented by Vargas et al. (2019) & Su et al.(2019).
-
```
beat-buesser, line 63: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -60,42 +60,46 @@
...
+        th: Optional[int] = None,
+        es: int = 0,
+        max_iter: int = 100,
+        targeted: bool = False,
         verbose: bool = True,
+        verbose_es: bool = False,
     ) -> None:
         """
         Create a :class:`.PixelThreshold` instance.
-
```
beat-buesser, line 84: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -60,42 +60,46 @@
...
         :param verbose: Print verbose messages of ES and show progress bars.
         """
         super().__init__(estimator=classifier)
 
         self._project = True
         self.type_attack = -1
-        self.th = th  # pylint: disable=C0103
-        self.es = es  # pylint: disable=C0103
+        self.th = th
+        self.es = es
```
beat-buesser, line 97:
```suggestion
        self.th = th  # pylint: disable=C0103
        self.es = es  # pylint: disable=C0103
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -60,42 +60,46 @@
...
+        self.max_iter = max_iter
         self._targeted = targeted
         self.verbose = verbose
+        self.verbose_es = verbose_es
+        self.rescale = False
```
beat-buesser, line 102:
```suggestion
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +125,12 @@ def _check_params(self) -> None:
...
+        if not isinstance(self.verbose_es, bool):
             raise ValueError("The argument `verbose` has to be of type bool.")
 
-    def generate(  # pylint: disable=W0221
-        self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int = 100, **kwargs
-    ) -> np.ndarray:
+    def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> np.ndarray:
         """
         Generate adversarial samples and return them in an array.
-
```
beat-buesser, line 132: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
beat-buesser: I would make it more explicit like:
```suggestion
        if self.estimator.clip_values[1] != 255.0:
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -149,47 +150,69 @@ def generate(  # pylint: disable=W0221
...
+
+        # NOTE: Pixel and Threshold Attacks are well defined for unprocessed images where the pixel values are,
+        #       8-Bit color i.e., the pixel values are np.uint8 in range [0, 255].
 
-        scale_input = bool(np.max(x) <= 1)
+        # TO-DO: Better checking of input image.
+        #        All other cases not tested needs the images to be rescaled to [0, 255].
+        if np.max(x) <= 1:
+            self.rescale = True
+            x = x * 255
```
beat-buesser, line 166:
```suggestion
            x = x / self.estimator.clip_values[1] * 255.0
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -149,47 +150,69 @@ def generate(  # pylint: disable=W0221
...
 
-        if scale_input:
-            adv_x_best_array = adv_x_best_array / 255.0
-
         if y is not None:
             y = to_categorical(y, self.estimator.nb_classes)
 
+        if self.rescale:
+            x = x.astype(np.float32) / 255.0
+            adv_x_best_array = adv_x_best_array.astype(np.float32) / 255.0
```
beat-buesser, line 214:
```suggestion
            x = x.astype(np.float32) / 255.0 * self.estimator.clip_values[1]
            adv_x_best_array = adv_x_best_array.astype(np.float32) / 255.0 * self.estimator.clip_values[1]
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -230,23 +253,31 @@ def _attack_success(self, adv_x, x, target_class):
         """
         Checks whether the given perturbation `adv_x` for the image `img` is successful.
         """
-        predicted_class = np.argmax(self.estimator.predict(self._perturb_image(adv_x, x))[0])
+        adv = self._perturb_image(adv_x, x)
+
+        if self.rescale:
+            adv = adv.astype(np.float32) / 255.0
```
beat-buesser, line 259:
```suggestion
            adv = adv.astype(np.float32) / 255.0 * self.estimator.clip_values[1]
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -230,23 +253,31 @@ def _attack_success(self, adv_x, x, target_class):
...
         untargeted attack and targeted label for targeted attack.
         """
         bounds, initial = self._get_bounds(image, limit)
 
         def predict_fn(x):
-            predictions = self.estimator.predict(self._perturb_image(x, image))[:, target_class]
+            adv = self._perturb_image(x, image)
+
+            if self.rescale:
+                adv = adv.astype(np.float32) / 255.0
```
beat-buesser, line 278:
```suggestion
                adv = adv.astype(np.float32) / 255.0 * self.estimator.clip_values[1]
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -448,9 +480,7 @@ def _perturb_image(self, x: np.ndarray, img: np.ndarray) -> np.ndarray:
 To speed up predictions, the entire parameters array is passed to `self.func`,
 where a neural network model can batch its computations and execute in parallel
 Search for `CHANGES` to find all code changes.
-
 Dan Kondratyuk 2018
-
```
beat-buesser, line 453: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -448,9 +480,7 @@ def _perturb_image(self, x: np.ndarray, img: np.ndarray) -> np.ndarray:
 To speed up predictions, the entire parameters array is passed to `self.func`,
 where a neural network model can batch its computations and execute in parallel
 Search for `CHANGES` to find all code changes.
-
```
beat-buesser, line 451: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -395,7 +427,6 @@ def _get_bounds(self, img: np.ndarray, limit) -> Tuple[List[list], list]:
 class ThresholdAttack(PixelThreshold):
     """
     This attack was originally implemented by Vargas et al. (2019).
-
```
beat-buesser, line 398: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -324,20 +355,21 @@ def __init__(
...
         th: Optional[int] = None,
-        es: int = 0,
+        es: int = 1,
+        max_iter: int = 100,
         targeted: bool = False,
         verbose: bool = False,
     ) -> None:
         """
         Create a :class:`.PixelAttack` instance.
-
```
beat-buesser, line 333: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -312,7 +344,6 @@ class PixelAttack(PixelThreshold):
     """
     This attack was originally implemented by Vargas et al. (2019). It is generalisation of One Pixel Attack originally
     implemented by Su et al. (2019).
-
```
beat-buesser, line 315: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
beat-buesser: I think we should add check for `clip_values` in `_check_params`

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +125,12 @@ def _check_params(self) -> None:
         if not isinstance(self.verbose, bool):
             raise ValueError("The flag `verbose` has to be of type bool.")
 
-        if not isinstance(self.verbose, bool):
+        if not isinstance(self.verbose_es, bool):
```
beat-buesser, line 128: Please add after this check the following check:
```
        if self.estimator.clip_values is None or self.estimator.clip_values[0] != 0:
            raise ValueError("This attack requires estimator clip values to be defined and the minimum value has to be 0.0.")
```

beat-buesser:
Hi @shashankkotyan Thank you very much! I have made a proposal to your question and added a few suggestions, most are related to the removed empty line in the docstrings, we need to add it back as it is required for the correct formatting of the docs website in html.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -282,18 +313,19 @@ def callback_fn(x, convergence=None):  # pylint: disable=R1710,W0613
...
                     maxfun=max(1, 400 // len(bounds)) * len(bounds) * 100,
                     callback=callback_fn,
-                    iterations=max_iter,
+                    iterations=self.max_iter,
                 )
-            except Exception as exception:  # pylint: disable=W0703
-                logger.info(exception)
+            except Exception as exception:
+                if self.verbose_es:
+                    print(exception)
```
beat-buesser, line 320: The exception's stack trace should be printed anyways.
```suggestion
            except Exception as exception:  # pylint: disable=W0703
                if self.verbose_es:
                    logger.exception("CMAEvolutionStrategy optimize failed.")
```

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Pylint

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Not required in the documentation.

art/attacks/evasion/pixel_threshold.py
shashankkotyan: Actually Most of the time, it would be informative and not a real exception handling.
CMA-ES docs recommend raising an exception to early stop optimisation. 
To do this, an exception is raised at line 288 in callback_fn, this line will print the same exception.

However, if you think, this workaround can be done better. I am open to suggestions.

art/attacks/evasion/pixel_threshold.py
```diff
@@ -405,19 +450,20 @@ def __init__(
         classifier: "CLASSIFIER_NEURALNETWORK_TYPE",
         th: Optional[int] = None,
         es: int = 0,
+        max_iter: int = 100,
         targeted: bool = False,
         verbose: bool = False,
     ) -> None:
         """
         Create a :class:`.PixelThreshold` instance.
-
```
beat-buesser, line 413: Please don't remove this line, it's needed for correct formatting of the docs.

art/attacks/evasion/pixel_threshold.py
beat-buesser: Ok, I see. How about defining a custom exception, that would also avoid the requirement for `# pylint: disable=W0703`? How about adding
 ```
class CMAEarlyStoppingException(Exception):
    """Raised when CMA is stopping early after successful optimisation."""
    pass
```
to this file?

and changing here to 
```suggestion
            except CMAEarlyStoppingException as e:
                if self.verbose_es:
                    logger.info(e.message)
```

beat-buesser: And change line 297 from
```
raise Exception("Attack Completed :) Earlier than expected")
```
to
```
raise CMAEarlyStoppingException("Attack Completed :) Earlier than expected")
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -60,31 +61,34 @@
 class PixelThreshold(EvasionAttack):
     """
     These attacks were originally implemented by Vargas et al. (2019) & Su et al.(2019).
-
+
```
beat-buesser, line 64:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -60,31 +61,34 @@
...
+        es: int = 0,
+        max_iter: int = 100,
+        targeted: bool = False,
         verbose: bool = True,
+        verbose_es: bool = False,
     ) -> None:
         """
         Create a :class:`.PixelThreshold` instance.
-
+
```
beat-buesser, line 87:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +127,21 @@ def _check_params(self) -> None:
...
             raise ValueError("The flag `verbose` has to be of type bool.")
 
-        if not isinstance(self.verbose, bool):
+        if not isinstance(self.verbose_es, bool):
             raise ValueError("The argument `verbose` has to be of type bool.")
-
-    def generate(  # pylint: disable=W0221
-        self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int = 100, **kwargs
-    ) -> np.ndarray:
+
```
beat-buesser, line 132:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +127,21 @@ def _check_params(self) -> None:
...
 
-        if not isinstance(self.verbose, bool):
+        if not isinstance(self.verbose_es, bool):
             raise ValueError("The argument `verbose` has to be of type bool.")
-
-    def generate(  # pylint: disable=W0221
-        self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int = 100, **kwargs
-    ) -> np.ndarray:
+       
+        if self.estimator.clip_values is None:
```
beat-buesser, line 133:
```suggestion
        if self.estimator.clip_values is None:
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +127,21 @@ def _check_params(self) -> None:
...
+        if not isinstance(self.verbose_es, bool):
             raise ValueError("The argument `verbose` has to be of type bool.")
-
-    def generate(  # pylint: disable=W0221
-        self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int = 100, **kwargs
-    ) -> np.ndarray:
+       
+        if self.estimator.clip_values is None: 
+            raise ValueError("This attack requires estimator clip values to be defined.")
+
```
beat-buesser, line 135:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +127,21 @@ def _check_params(self) -> None:
...
-    ) -> np.ndarray:
+       
+        if self.estimator.clip_values is None: 
+            raise ValueError("This attack requires estimator clip values to be defined.")
+    
+    def rescale_input(self, x):
+        x = x.astype(np.float32) / 255.0
+        x = (x * (self.estimator.clip_values[1] - self.estimator.clip_values[0])) + self.estimator.clip_values[0]
+        return x
+
```
beat-buesser, line 140:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +127,21 @@ def _check_params(self) -> None:
...
+    def rescale_input(self, x):
+        x = x.astype(np.float32) / 255.0
+        x = (x * (self.estimator.clip_values[1] - self.estimator.clip_values[0])) + self.estimator.clip_values[0]
+        return x
+        
+    def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> np.ndarray:
         """
         Generate adversarial samples and return them in an array.
-
+
```
beat-buesser, line 144:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -149,47 +161,70 @@ def generate(  # pylint: disable=W0221
...
 
-        if scale_input:
-            adv_x_best_array = adv_x_best_array / 255.0
-
         if y is not None:
             y = to_categorical(y, self.estimator.nb_classes)
 
+        if self.rescale:
+            x = self.rescale_input(x)
+            adv_x_best_array =  self.rescale_input(adv_x_best_array)
```
beat-buesser, line 226:
```suggestion
            adv_x_best_array = self.rescale_input(adv_x_best_array)
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -312,7 +356,7 @@ class PixelAttack(PixelThreshold):
     """
     This attack was originally implemented by Vargas et al. (2019). It is generalisation of One Pixel Attack originally
     implemented by Su et al. (2019).
-
+
```
beat-buesser, line 359:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -324,20 +368,22 @@ def __init__(
...
-        es: int = 0,
+        es: int = 1,
+        max_iter: int = 100,
         targeted: bool = False,
         verbose: bool = False,
     ) -> None:
         """
         Create a :class:`.PixelAttack` instance.
-
+
```
beat-buesser, line 378:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -395,7 +441,7 @@ def _get_bounds(self, img: np.ndarray, limit) -> Tuple[List[list], list]:
 class ThresholdAttack(PixelThreshold):
     """
     This attack was originally implemented by Vargas et al. (2019).
-
+
```
beat-buesser, line 444:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -405,19 +451,21 @@ def __init__(
...
         th: Optional[int] = None,
         es: int = 0,
+        max_iter: int = 100,
         targeted: bool = False,
         verbose: bool = False,
     ) -> None:
         """
         Create a :class:`.PixelThreshold` instance.
-
+
```
beat-buesser, line 460:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -121,15 +125,17 @@ def _check_params(self) -> None:
...
-    def generate(  # pylint: disable=W0221
-        self, x: np.ndarray, y: Optional[np.ndarray] = None, max_iter: int = 100, **kwargs
-    ) -> np.ndarray:
+        if self.estimator.clip_values is None:
+            raise ValueError("This attack requires estimator clip values to be defined.")
+    def rescale_input(self, x):
+        x = x.astype(np.float32) / 255.0
+        x = (x * (self.estimator.clip_values[1] - self.estimator.clip_values[0])) + self.estimator.clip_values[0]
+        return x
+    def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> np.ndarray:
```
beat-buesser, line 136:
```suggestion

    def rescale_input(self, x):
        x = x.astype(np.float32) / 255.0
        x = (x * (self.estimator.clip_values[1] - self.estimator.clip_values[0])) + self.estimator.clip_values[0]
        return x

    def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> np.ndarray:
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -438,7 +476,10 @@ def _perturb_image(self, x: np.ndarray, img: np.ndarray) -> np.ndarray:
             ):
                 image[i, j, k] = adv[count]
         return imgs
-
+
```
beat-buesser, line 479:
```suggestion

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -438,7 +476,10 @@ def _perturb_image(self, x: np.ndarray, img: np.ndarray) -> np.ndarray:
             ):
                 image[i, j, k] = adv[count]
         return imgs
-
+    
+class CMAEarlyStoppingException(Exception):
+    """Raised when CMA is stopping early after successful optimisation."""
+    pass
```
beat-buesser, line 482:
```suggestion
    pass

```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -282,18 +319,19 @@ def callback_fn(x, convergence=None):  # pylint: disable=R1710,W0613
...
                     maxfun=max(1, 400 // len(bounds)) * len(bounds) * 100,
                     callback=callback_fn,
-                    iterations=max_iter,
+                    iterations=self.max_iter,
                 )
-            except Exception as exception:  # pylint: disable=W0703
-                logger.info(exception)
+            except CMAEarlyStoppingException as e:
+                if self.verbose_es:
+                    logger.info(e.message)
```
beat-buesser, line 326:
```suggestion
                    logger.info(e)
```

beat-buesser, line 326:
```suggestion
            except CMAEarlyStoppingException as err:
                if self.verbose_es:
                    logger.info(err)
```

art/attacks/evasion/pixel_threshold.py
```diff
@@ -440,6 +481,11 @@ def _perturb_image(self, x: np.ndarray, img: np.ndarray) -> np.ndarray:
         return imgs
 
 
+class CMAEarlyStoppingException(Exception):
+    """Raised when CMA is stopping early after successful optimisation."""
+    pass
```
beat-buesser, line 486:
```suggestion
    """Raised when CMA is stopping early after successful optimisation."""

    pass
```

beat-buesser (approved):
Hi @shashankkotyan Thank you very much for your updates to Pixel Attack and Threshold Attack!

shashankkotyan:
Hi @beat-buesser!

There was one more minor change, I wanted to do... 
In [Line 177 ](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/86baa09ffeaf08c2b2e1ef75fa80af5ba3f8ff74/art/attacks/evasion/pixel_threshold.py#L177) of the merged update, I did `x = x.astype(np.uint8)`, however I noticed that in other attacks the same operation were done as ` x = x.astype(ART_NUMPY_DTYPE)`. 

So if possible, can you change the particular line to be more coherent to the library and other attacks.

Apologies for the inconvenience caused.

PR #1052, opened 2021-04-15
Generalize DeepSpeech estimator and asr attack to support espresso

minhitbk:
# Description
This PR generalizes codes to support espresso estimator.

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:
- OS
- Python version

# Checklist


art/estimators/speech_recognition/pytorch_deep_speech.py
```diff
@@ -676,6 +724,43 @@ def _transform_model_input(
...
+        """
+        self.model.train()
+
+    def batch_norm(self, train: bool) -> None:
+        """
+        Set all batch normalization layers into train or eval mode.
+
+        :param train: False for evaluation mode.
+        """
+        self.set_batchnorm(train=train)
```
beat-buesser, line 762: What is the reason for this method? It just seems to call the existing method.

beat-buesser (changes requested):
Hi @minhitbk Thank you very much for your pull request. I think it looks good. I have only one question. What do you think?

YiwenShaoStephen:
Hi @minhitbk Thanks for this great update which makes imperceptible attack work more generally for other estimators.
I just have a confusion that do you think the windowing parameters used in this attack should be tied with those used in estimators? 
It seems to me that these windowing parameters can be fixed with the attack itself since it's targeting on Masking effects for spectrogram and could be a general function for audio samples without involving with the exact model. 
And also in practice, the feature extractors (e.g. fbank, stft, mfcc) may vary in different speech recognition models (some of them might don't have a one) so it would be very tricky if we want to support other models in the future.
Please correct me if I misunderstanding something here.

art/estimators/speech_recognition/pytorch_deep_speech.py
minhitbk: @beat-buesser : We need to call this in the attack for the Deepspeech estimator, but the Espresso one does not need to call set_batchnorm, so they need to provide an empty method. Please note that the set_batchnorm is inherited from the PytorchEstimator where Espresso also inherits, so we cannot call this method directly.

minhitbk:
@YiwenShaoStephen : Thanks very much for your comment. I agree with you that those window params should be part of the attack instead of the estimator. I discussed with @beat-buesser , so we will implement those params as attack's attributes which are passed to the attack through its `__init__` function. In the meanwhile, you can assume that you could get those from attack's self and don't need to wait us. We will update this PR soon. Thanks.

beat-buesser (approved):
Hi @minhitbk Thank you very much!

PR #1081, opened 2021-05-04
Development issue 1062 update

yusong-tan:
# Description

1. Added a request to generalize rotation to work with Pytorch and TF box formats
2. Fixed code where brightness adjustment was inconsistent with input and patch assumptions
3. Commented on whether or not "brightness_range[1] <= 1.0" is a necessary requirement.

I had initially planned to add arbitrary perspective transforms as an additional EOT function and as another way to insert patch but decided against it because it required each batch to contain highly correlated images.


beat-buesser (approved):
Hi @yusong-tan Thank you!

PR #1175, opened 2021-06-17
Minor bug of Dtype casting in Pixel and Threshold Attack

shashankkotyan:
Fixes Dtype Casting in Pixel and Threshold Attack.


beat-buesser (approved):
Hi @shashankkotyan Thank you very much for these updates!

PR #1192, opened 2021-06-26
norms per sample; and roll left, right over frames, not batch

mbc:
# Description

This fixes two bugs in the flicker attack. The first is that the regularization norms are computed over the whole batch of perturbations, rather than per sample's perturbation. The second is that the "roll" operations are performed over the batch samples, rather than over the frames.

In practice, for a batch size of 1 (common for videos), only the second bug gets activated; but it essentially disables the regularization.

## Type of change

- [X] Bug fix (non-breaking)

# Testing

I tested that, prior to the bug fix `beta_0` has essentially no effect, across a range of orders of magnitude, on a few UCF101 samples. I then tested that, after the bug fix, a large enough `beta_0` will force the attack to fail, while a small enough `beta_0` will allow it to succeed (again over a few UCF101 samples).

# Checklist


beat-buesser (approved):
Hiu @mbc Thank you very much for discovering and fixing these bugs!

PR #1197, opened 2021-06-29
New boundary attack

### LINKED ISSUE #1167: Implement Boundary Attack from Li and Zhang
Would be interesting to compare to Choquette-Choo et al.'s work which is already implemented (`LabelOnlyDecisionBoundary`).

Both attacks are basically the same, except for how they calibrate the distance threshold tau (which is used to infer membership). Therefore I propose to extend the existing `LabelOnlyDecisionBoundary` with a new threshold calibrating function. Thus we will have two functions besides `infer(x,y)`:
- `calibrate_distance_threshold(x_train, y_train, x_test, y_test)` - Choquette-Choo et al. distance threshold calibration (already implemented)
- **`calibrate_distance_threshold_unsupervised(top_t, num_samples, max_queries)`** - Li and Zhang distance threshold calibration
    - naming: it is `unsupervised` because it requires no data to set the distance threshold
    - Implements Algorithm 2 from the appendix of the paper. Input:
            - `top_t`: top-t percentile
            - `num_samples`: number of random samples to generate
            - `max_queries`: maximum number of queries
 
I have considered inheritance from `LabelOnlyDecisionBoundary`, however this would require me to either:
- override `calibrate_distance_threshold()` in the child class with the parameters shown above (which if I know correctly is bad practice)
- pass the above parameters upon creation of the attack in `init()`, and call `calibrate_distance_threshold()` with no passed parameters (would require change of signature in parent method with default values set to `None` + parameter check in parent so it doesn't get called with `None`)

**Additional context**
Paper link: https://arxiv.org/abs/2007.15528

minaremeli:
# Description

This is an implementation of a new boundary attack by Li and Zhang (https://arxiv.org/abs/2007.15528).

Fixes #1167

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

I included a test in `tests/attacks/inference/membership_inference/test_label_only_boundary_distance.py` which runs the newly implemented calibration method (`calibrate_distance_threshold_unsupervised`). This test is basically a copy of `test_label_only_boundary_distance_prob_calib`.
- [x] `test_label_only_boundary_distance_prob_calib_unsup`

**Test Configuration**:
- Ubuntu 20.04 (Focal Fossa)
- Python 3.8.10
- commit number: `4c5b12eb`
- TensorFlow  2.3.2 / Keras 2.4.3 / PyTorch 1.8.1 / MXNet (not installed)

# Checklist

### DIFF
--- a/art/attacks/inference/membership_inference/label_only_boundary_distance.py
+++ b/art/attacks/inference/membership_inference/label_only_boundary_distance.py
@@ -207,7 +207,7 @@ class LabelOnlyDecisionBoundary(MembershipInferenceAttack):
     ):
         """
         Calibrate distance threshold on randomly generated samples, choosing the top-t percentile of the noise needed
-        to change the classifier's initial prediction.
+        to change the classifier's initial prediction. This method requires the model's clip_values to be set.
 
         | Paper link: https://arxiv.org/abs/2007.15528
 
--- a/art/attacks/inference/membership_inference/label_only_boundary_distance.py
+++ b/art/attacks/inference/membership_inference/label_only_boundary_distance.py
@@ -18,7 +18,8 @@
 """
 This module implements the Label-Only Inference Attack based on Decision Boundary.
 
-| Paper link: https://arxiv.org/abs/2007.14321
+| Paper link: https://arxiv.org/abs/2007.14321 (Choquette-Choo et al.) and https://arxiv.org/abs/2007.15528 (Li and
+Zhang)
 """
 import logging
 from typing import Optional, TYPE_CHECKING
@@ -40,7 +41,10 @@ class LabelOnlyDecisionBoundary(MembershipInferenceAttack):
     """
     Implementation of Label-Only Inference Attack based on Decision Boundary.
 
-    | Paper link: https://arxiv.org/abs/2007.14321
+    | Paper link: https://arxiv.org/abs/2007.14321 (Choquette-Choo et al.) and https://arxiv.org/abs/2007.15528 (Li
+    and Zhang)
+
+    You only need to call ONE of the calibrate methods, depending on which attack you want to launch.
     """
 
     attack_params = MembershipInferenceAttack.attack_params + [
@@ -135,6 +139,8 @@ class LabelOnlyDecisionBoundary(MembershipInferenceAttack):
     ):
         """
         Calibrate distance threshold maximising the membership inference accuracy on `x_train` and `x_test`.
+
+        | Paper link: https://arxiv.org/abs/2007.14321
 
         :param x_train: Training data.
         :param y_train: Labels of training data `x_train`.
--- a/art/attacks/inference/membership_inference/label_only_boundary_distance.py
+++ b/art/attacks/inference/membership_inference/label_only_boundary_distance.py
@@ -202,7 +202,9 @@ class LabelOnlyDecisionBoundary(MembershipInferenceAttack):
 
         self.distance_threshold_tau = distance_threshold_tau
 
-    def calibrate_distance_threshold_unsupervised(self, top_t: int, num_samples: int, max_queries: int, **kwargs):
+    def calibrate_distance_threshold_unsupervised(
+        self, top_t: int = 50, num_samples: int = 100, max_queries: int = 1, **kwargs
+    ):
         """
         Calibrate distance threshold on randomly generated samples, choosing the top-t percentile of the noise needed
         to change the classifier's initial prediction.
@@ -211,7 +213,8 @@ class LabelOnlyDecisionBoundary(MembershipInferenceAttack):
 
         :param top_t: Top-t percentile.
         :param num_samples: Number of random samples to generate.
-        :param max_queries: Maximum number of queries.
+        :param max_queries: Maximum number of queries. Maximum number of HSJ iterations on a single sample will be
+                            max_queries * max_iter.
         :Keyword Arguments for HopSkipJump:
             * *norm*: Order of the norm. Possible values: "inf", np.inf or 2.
             * *max_iter*: Maximum number of iterations.
@@ -222,13 +225,18 @@ class LabelOnlyDecisionBoundary(MembershipInferenceAttack):
         """
         from art.attacks.evasion.hop_skip_jump import HopSkipJump
 
-        x_min, x_max = self.estimator.clip_values
+        if self.estimator.clip_values is not None:
+            x_min, x_max = self.estimator.clip_values
+        else:
+            raise RuntimeError(
+                "You need to set the estimator's clip_values in order to calibrate the distance threshold."
+            )
 
         x_rand = np.random.rand(*(num_samples,) + self.estimator.input_shape).astype(np.float32)
         x_rand *= x_max - x_min  # scale
         x_rand += x_min  # shift
 
-        y_rand = np.random.randint(0, self.estimator.nb_classes, num_samples)
+        y_rand = self.estimator.predict(x=x_rand)
         y_rand = check_and_transform_label_format(y_rand, self.estimator.nb_classes)
 
         hsj = HopSkipJump(classifier=self.estimator, targeted=False, **kwargs)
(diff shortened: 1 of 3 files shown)


minaremeli:
I only ran existing tests from `test_label_only_boundary_distance.py` since my changes only affected the `LabelOnlyDecisionBoundary` attack. Should I run all tests just in case?

beat-buesser:
I think that should be ok. I just started the full set of checks here which I expect should pass.

art/attacks/inference/membership_inference/label_only_boundary_distance.py
```diff
@@ -196,6 +202,62 @@ def calibrate_distance_threshold(
 
         self.distance_threshold_tau = distance_threshold_tau
 
+    def calibrate_distance_threshold_unsupervised(self, top_t: int, num_samples: int, max_queries: int, **kwargs):
```
abigailgold, line 205: Is it possible to provide default values for the params top_t, num_samples and max_queries?

art/attacks/inference/membership_inference/label_only_boundary_distance.py
```diff
@@ -196,6 +202,62 @@ def calibrate_distance_threshold(
...
+            * *norm*: Order of the norm. Possible values: "inf", np.inf or 2.
+            * *max_iter*: Maximum number of iterations.
+            * *max_eval*: Maximum number of evaluations for estimating gradient.
+            * *init_eval*: Initial number of evaluations for estimating gradient.
+            * *init_size*: Maximum number of trials for initial generation of adversarial examples.
+            * *verbose*: Show progress bars.
+        """
+        from art.attacks.evasion.hop_skip_jump import HopSkipJump
+
+        x_min, x_max = self.estimator.clip_values
```
abigailgold, line 225: Need some solution for the case that estimator.clip_values is not set (it is an optional param when initializing an estimator).

art/attacks/inference/membership_inference/label_only_boundary_distance.py
```diff
@@ -196,6 +202,62 @@ def calibrate_distance_threshold(
...
+    def calibrate_distance_threshold_unsupervised(self, top_t: int, num_samples: int, max_queries: int, **kwargs):
+        """
+        Calibrate distance threshold on randomly generated samples, choosing the top-t percentile of the noise needed
+        to change the classifier's initial prediction.
+
+        | Paper link: https://arxiv.org/abs/2007.15528
+
+        :param top_t: Top-t percentile.
+        :param num_samples: Number of random samples to generate.
+        :param max_queries: Maximum number of queries.
```
abigailgold, line 214: Isn't this number actually multiplied by the num_queries param supplied to hsj? I think this should be clarified...

art/attacks/inference/membership_inference/label_only_boundary_distance.py
minaremeli: Of course! I would suggest top_t to be set to 50% (because in the paper the best attack performances were achieved between 30% and 80%). 
num_samples could be 100 for example? And max_queries 1, so it doesn't run forever. But these values are pretty much ad-hoc.

art/attacks/inference/membership_inference/label_only_boundary_distance.py
minaremeli: If there was an input array `x` passed to the function I would do a similar approach to the hsj implementation (but there isn't :crying_cat_face: ):
```python
# Get clip_min and clip_max from the classifier or infer them from data
if self.estimator.clip_values is not None:
    clip_min, clip_max = self.estimator.clip_values
else:
    clip_min, clip_max = np.min(x), np.max(x)
```
Another solution would be to do a check at the beginning of the function and throw an error if clip_values is not set.

Or maybe the best solution would be to do a check, and if it's None then set default values 0 and 1 and throw a warning. Something like this:
```python
if self.estimator.clip_values is not None:
    x_min, x_max = self.estimator.clip_values
else:
    x_min, x_max = 0, 1
    raise RuntimeWarning("The estimator has no clip_values set. Generated feature values will range between 0 and 1.")
```

art/attacks/inference/membership_inference/label_only_boundary_distance.py
minaremeli: Do you mean the `max_iter` parameter of hsj? You are right, the maximum number of iterations one does with hsj on a given sample is `max_queries * max_iter`. Do you want me to clarify this?

art/attacks/inference/membership_inference/label_only_boundary_distance.py
```diff
@@ -196,6 +202,62 @@ def calibrate_distance_threshold(
...
+        """
+        from art.attacks.evasion.hop_skip_jump import HopSkipJump
+
+        x_min, x_max = self.estimator.clip_values
+
+        x_rand = np.random.rand(*(num_samples,) + self.estimator.input_shape).astype(np.float32)
+        x_rand *= x_max - x_min  # scale
+        x_rand += x_min  # shift
+
+        y_rand = np.random.randint(0, self.estimator.nb_classes, num_samples)
```
minaremeli, line 231: I found a mistake - `y_rand` should be the predicted label given `x_rand` (according to Algorithm 3 of the paper). Should be something like this:
```python
y_rand = self.estimator.predict(x=x_rand)
```

art/attacks/inference/membership_inference/label_only_boundary_distance.py
abigailgold: Yes, this is what I meant. You can just add another sentence to the end of the parameter description.

art/attacks/inference/membership_inference/label_only_boundary_distance.py
abigailgold: That's fine. The user can always override them, but it's good to have the option (especially for non-experts) to run it without having to supply anything.

art/attacks/inference/membership_inference/label_only_boundary_distance.py
abigailgold: Will the calibration work properly even if wrong (default) min and max values are used? Or is this critical for the correctness of the threshold? If it's the latter, I prefer to raise an exception instead of using default values so that the user knows that these must be set. @beat-buesser what are your thoughts on this?

art/attacks/inference/membership_inference/label_only_boundary_distance.py
beat-buesser: Hi @minaremeli Are you planning to add an additional commit for this item?

minaremeli: Hi @beat-buesser sorry for the delay - yes I plan to add an additional commit as soon as possible (this week).

beat-buesser: Great, no worries!

art/attacks/inference/membership_inference/label_only_boundary_distance.py
minaremeli: I changed it to raise a `RuntimeError` in the newest commit ([a0e924c](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/1197/commits/a0e924c751a4ee47f34e5b06a3aef66c07651772)). Is this ok @beat-buesser ?

art/attacks/inference/membership_inference/label_only_boundary_distance.py
```diff
@@ -196,6 +202,70 @@ def calibrate_distance_threshold(
 
         self.distance_threshold_tau = distance_threshold_tau
 
+    def calibrate_distance_threshold_unsupervised(
+        self, top_t: int = 50, num_samples: int = 100, max_queries: int = 1, **kwargs
+    ):
+        """
+        Calibrate distance threshold on randomly generated samples, choosing the top-t percentile of the noise needed
+        to change the classifier's initial prediction.
```
abigailgold, line 210: Can you please add to the method description a statement that says that it requires the model's clip_values to be set? Thanks.

minaremeli: Done!

beat-buesser (approved):
Hi @minaremeli Thank you very much for your great contribution to ART enabling label only membership inference attacks without access to correct training labels and providing a very informative notebook that will help many users of ART getting started with these new inference tools. We hope you enjoy working on ART and are looking forward to many more contributions!

PR #1262, opened 2021-08-11
Fix min_epsilon of BoundaryAttack to receive a non-negative value

### LINKED ISSUE #1258: Exception is raised even if min_epsilon is a positive float in BoundaryAttack
**Describe the bug**
[`BoundaryAttack`](https://adversarial-robustness-toolbox.readthedocs.io/en/latest/modules/attacks/evasion.html#boundary-attack-decision-based-attack) cannot receive `min_epsilon`.

**To Reproduce**
1. Train a model.
2. Call `BoundaryAttack(model, min_epsilon=0.001)`.
3. See an exception.

**Expected behavior**
`min_epsilon` is a positive `float` (or `None`), so `0.001` should be valid.

**Cause of the bug**  
[boundary.py:L473-L474](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/attacks/evasion/boundary.py#L473-L474)  
```python
if self.min_epsilon is not None and (isinstance(self.min_epsilon, float) or self.min_epsilon <= 0):
    raise ValueError("The minimum epsilon must be a positive float.")
```
raise error when `self.min_epsilon` is `float`. Therefore, we need `not` before the `isinstance`.

kaitokishi:
Signed-off-by: kaitokishi <kishi.kaito@fujitsu.com>

# Description

Passing a positive float value into `min_epsilon` of `BoundaryAttack` raises Exception. This is because `_check_params()` raises `ValueError` even if `self.min_epsilon` is positive float. This PR is to fix this bug.

Also, this PR makes it simple to use `min_epsilon` by avoiding to use `None`.

Fixes #1258 

## Type of change

- [x] Bug fix (non-breaking)

# Testing

- [x] Below code
```
attack = BoundaryAttack(model, batch_size=100, targeted=False, max_iter=20, epsilon=0.1, min_epsilon=0.02)
x_adv = attack.generate(x=x)
```
works.

**Test Configuration**:
- OS: Ubuntu 18.04.5 LTS
- Python version: 3.6.9
- ART version or commit number: 1.7.0
- TensorFlow version: 2.3.0

# Checklist

### DIFF
--- a/art/attacks/evasion/boundary.py
+++ b/art/attacks/evasion/boundary.py
@@ -470,7 +470,7 @@ class BoundaryAttack(EvasionAttack):
         if self.step_adapt <= 0 or self.step_adapt >= 1:
             raise ValueError("The adaptation factor must be in the range (0, 1).")
 
-        if self.min_epsilon < 0:
+        if not isinstance(self.min_epsilon, (float, int)) or self.min_epsilon < 0:
             raise ValueError("The minimum epsilon must be non-negative.")
 
         if not isinstance(self.verbose, bool):
--- a/art/attacks/evasion/boundary.py
+++ b/art/attacks/evasion/boundary.py
@@ -76,7 +76,7 @@ class BoundaryAttack(EvasionAttack):
         num_trial: int = 25,
         sample_size: int = 20,
         init_size: int = 100,
-        min_epsilon: Optional[float] = None,
+        min_epsilon: float = 0,
         verbose: bool = True,
     ) -> None:
         """
@@ -327,7 +327,7 @@ class BoundaryAttack(EvasionAttack):
                 logger.warning("Adversarial example found but not optimal.")
                 return self._best_adv(original_sample, x_advs)
 
-            if self.min_epsilon is not None and self.curr_epsilon < self.min_epsilon:
+            if self.curr_epsilon < self.min_epsilon:
                 return x_adv
 
         return x_adv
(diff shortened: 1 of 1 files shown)


kaitokishi:
I would like to make a pull request to `dev_1.7.2` but I cannot select the branch because there is no `dev_1.7.2` branch. Thus, I selected `dev_1.8.0`.

art/attacks/evasion/boundary.py
```diff
@@ -76,7 +76,7 @@ def __init__(
         num_trial: int = 25,
         sample_size: int = 20,
         init_size: int = 100,
-        min_epsilon: Optional[float] = None,
+        min_epsilon: float = 0,
```
beat-buesser, line 79:
```suggestion
        min_epsilon: Union[float, int] = 0,
```

beat-buesser (changes requested):
Hi @kaitokishi Thank you very much for your first pull request to ART! This is really great! I have created branch `dev_1.7.2` now and updated the PR accordingly. I like removing the `None`, but I would like to suggest to apply typing and testing for `int` and `float` explicitly for good testing. What do you think?

art/attacks/evasion/boundary.py
kaitokishi: In [PEP 484](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower),
>when an argument is annotated as having type float, an argument of type int is acceptable

so I think only `float` is sufficient as in `epsilon` and `delta`.

kaitokishi:
I think using `Union[float, int]` is redundent as I commented. Nevertheless, if you would like to write it explicitly, I accept your suggestion.

beat-buesser (approved):
Hi @kaitokishi Thank you very much for you first pull request to ART!

PR #1380, opened 2021-10-24
Updates for CarliniLInfMethod attack

### LINKED ISSUE #1374: CarliniLInfMethod differs significantly from the original, resulting in lower attack success rate
**Describe the bug**  
The loss used in `CarliniLInfMethod` differs from [the paper](https://arxiv.org/pdf/1608.04644.pdf). Therefore, `CarliniLInfMethod` fails to generate adversarial images in small `eps` while [the original implementation by Carlini](https://github.com/carlini/nn_robust_attacks/blob/master/li_attack.py) can generate adversarial images in the `eps` perturbation size.

**To reproduce**  
1. Make a simple neural network to classify MNIST.
2. Generate adversarial images by [the original implementation](https://github.com/carlini/nn_robust_attacks/blob/master/li_attack.py). We can get the images in small perturbation sizes like `0.1`.
3. Generate adversarial images by `CarliniLInfMethod` with `eps=0.1` which is sufficient in the original implementation, and get lower attack success rate. I tried random 10 samples from MNIST and got 100% attack success rate (i.e. 10 samples out of 10) in the original implementation but got 60% attack success rate (i.e. 6 samples out of 10, which are the same as used in the original implementation) in `CarliniLInfMethod`.

**Expected behavior**  
`CarliniLInfMethod` should have high attack success rate comparable with Carlini's original implementation even if `eps` is small as in the original implementation.

**Cause of the bug**  
The loss of Carlini & Wagner </img> is </img> (see p.10 of [the paper](https://arxiv.org/pdf/1608.04644.pdf)) while the loss implemented on `CarliniLInfMethod` is only </img>, so they are inconsistent.

 - OS: Ubuntu 18.04.5 LTS
 - Python version: 3.6.9
 - ART version or commit number: 1.8.0
 - To test Carlini's original implementation:
   + tensorflow-gpu version: 1.14.0 / Keras version: 2.3.0
 - To test `CarliniLInfMethod`:
   + TensorFlow version: 2.3.0

beat-buesser:
# Description

This pull request updates `CarliniLInfMethod` by adding perturbation regularisation, creating a new module for optimizers and adding Adam optimizer to replace line search optimisation, and adding searches for optimal maximal perturbation `tau` and loss scaling constant `c` to reproduce performance of the original publication's implementation.

Fixes #1374

## Type of change

- [ ] Bug fix (non-breaking)

# Checklist


art/attacks/evasion/carlini.py
```diff
@@ -515,10 +516,10 @@ class CarliniLInfMethod(EvasionAttack):
         "targeted",
         "learning_rate",
         "max_iter",
-        "max_halving",
-        "max_doubling",
-        "eps",
-        "batch_size",
+        "decrease_factor",
```
hesseltuinhof, line 519: If you change parameter names here, maybe also consider changing them for the L0 and L2 counterparts.

art/attacks/evasion/carlini.py
```diff
@@ -530,10 +531,10 @@ def __init__(
         targeted: bool = False,
         learning_rate: float = 0.01,
         max_iter: int = 10,
-        max_halving: int = 5,
-        max_doubling: int = 5,
-        eps: float = 0.3,
-        batch_size: int = 128,
```
hesseltuinhof, line 536: Is removal of `batch_size` intended?

art/attacks/evasion/carlini.py
```diff
@@ -546,9 +547,11 @@ def __init__(
...
         :param max_iter: The maximum number of iterations.
-        :param max_halving: Maximum number of halving steps in the line search optimization.
-        :param max_doubling: Maximum number of doubling steps in the line search optimization.
-        :param eps: An upper bound for the L_0 norm of the adversarial perturbation.
+        :param decrease_factor: The rate of shrinking tau, values in `0 < decrease_factor < 1` where larger is more
+                                accurate.
+        :param initial_const: The initial value of constant `c`.
+        :param largest_const: The largest value of constant `c`.
+        :param const_factor: The rate of increasing constant `c` with `const_factor > 1`, where smaller more accurate.
         :param batch_size: Size of the batch on which adversarial samples are generated.
```
hesseltuinhof, line 555: If removal of `batch_size` is intended, then this line can be removed as well.

art/attacks/evasion/carlini.py
```diff
@@ -558,26 +561,31 @@ def __init__(
...
+        :param x_adv: An array with the adversarial examples.
         :param target: An array with the target class (one-hot encoded).
-        :return: A tuple holding the current predictions and overall loss.
+        :param x: Benign samples.
+        :param  const: Current constant `c`.
+        :param tau: Current limit `tau`.
+        :return: A tuple of current predictions, total loss, logits loss and regularisation loss.
         """
-        z_predicted = self.estimator.predict(np.array(x_adv, dtype=ART_NUMPY_DTYPE), batch_size=self.batch_size)
+        z_predicted = self.estimator.predict(np.array(x_adv, dtype=ART_NUMPY_DTYPE), batch_size=1)
```
hesseltuinhof, line 588: Why is it necessary to hard-code a single prediction?

art/attacks/evasion/carlini.py
```diff
@@ -649,206 +736,92 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> n
...
+            clip_min, clip_max = np.amin(x), np.amax(x)
 
         # Assert that, if attack is targeted, y_val is provided:
         if self.targeted and y is None:
             raise ValueError("Target labels `y` need to be provided for a targeted attack.")
 
         # No labels provided, use model prediction as correct class
         if y is None:
-            y = get_labels_np_array(self.estimator.predict(x, batch_size=self.batch_size))
+            y = get_labels_np_array(self.estimator.predict(x, batch_size=1))
```
hesseltuinhof, line 749: See above

hesseltuinhof (changes requested):
I have a few questions. Otherwise, the changes look good :+1:

art/attacks/evasion/carlini.py
beat-buesser: Yes, this is intended because the new version follows closely the author's original implementation and does not work with mini-batches.

art/attacks/evasion/carlini.py
beat-buesser: I agree.

art/attacks/evasion/carlini.py
beat-buesser: I have set it to 1 because `x_adv` will always be a single sample and to express this explicitly.

art/attacks/evasion/carlini.py
beat-buesser: Same as above.

art/attacks/evasion/carlini.py
beat-buesser: Good point. I have considered the L2 method but it is based on a different algorithm/optimisation and requires different inputs.

art/attacks/evasion/carlini.py
hesseltuinhof: Thanks for the clarification. Given that it is not immediately clear why we choose `batch_size=1`, should we add a comment at an appropriate location in the code to mention why we explicitly use a batch size of one?

hesseltuinhof (approved):

PR #1422, opened 2021-11-22
Add support for multiple samples in SimBA.generate

### LINKED ISSUE #1407: change docstring for `SimBa.generate` parameter `y`
Currently the docstring for `Simba.generate` states that `y` should be class labels (as is true for most `art` attacks).

https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/85ca10ff6b696dbaa66dcf179d0493b12ff594bb/art/attacks/evasion/simba.py#L104

But actually in this case the `SimBA.generate` method expects a (1, c) matrix of class probabilities

I understand there's trade-offs when trying to provide a consistent API, but it's definitely confusing that the docstring disagrees with what the code expects.

Maybe keep the argument name but change the docstring?

* don't allow 'y' argument and just always use the `classifier` predictions directly
Not clear to me clear what's gained by allowing a user to do exactly the same thing as what the attack class does in this case.
I guess a user could provide different values for the probabilities if what the `classifier` gives is somehow not desired, say, rounded?

* Change the argument name to something else, say `y_probs`, to be completely explicit?
It's also not great when an argument with a consistent name has different semantics depending on the class.
Maybe all black box attacks could adopt this parameter.

Not sure if either of those is better, just putting alternatives out there.

beat-buesser:

# Description

This pull request adds support for multiple samples in `SimBA.generate`.

Fixes #1407

## Type of change

- [x] Bug fix (non-breaking)

# Checklist

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings

### DIFF
--- a/art/attacks/evasion/simba.py
+++ b/art/attacks/evasion/simba.py
@@ -32,6 +32,7 @@ from art.attacks.attack import EvasionAttack
 from art.estimators.estimator import BaseEstimator, NeuralNetworkMixin
 from art.estimators.classification.classifier import ClassifierMixin
 from art.config import ART_NUMPY_DTYPE
+from art.utils import is_probability
 
 if TYPE_CHECKING:
     from art.utils import CLASSIFIER_TYPE
@@ -74,7 +75,7 @@ class SimBA(EvasionAttack):
         """
         Create a SimBA (dct) attack instance.
 
-        :param classifier: A trained classifier.
+        :param classifier: A trained classifier predicting probabilities and not logits.
         :param attack: attack type: pixel (px) or DCT (dct) attacks
         :param max_iter: The maximum number of iterations.
         :param epsilon: Overshoot parameter.
@@ -107,9 +108,15 @@ class SimBA(EvasionAttack):
         x = x.astype(ART_NUMPY_DTYPE)
         x_adv = x.copy()
 
-        preds = self.estimator.predict(x, batch_size=self.batch_size)
+        y_prob_pred = self.estimator.predict(x, batch_size=self.batch_size)
 
-        if self.estimator.nb_classes == 2 and preds.shape[1] == 1:
+        if not is_probability(y_prob_pred):
+            raise ValueError(
+                "This attack requires an estimator predicting probabilities. It looks like the current "
+                "estimator is not predicting probabilities"
+            )
+
+        if self.estimator.nb_classes == 2 and y_prob_pred.shape[1] == 1:
             raise ValueError(
                 "This attack has not yet been tested for binary classification with a single output classifier."
             )
@@ -129,7 +136,7 @@ class SimBA(EvasionAttack):
 
             # Use model predictions as correct outputs
             logger.info("Using the model prediction as the correct label for SimBA.")
-            y_i = np.argmax(preds, axis=1)
+            y_i = np.argmax(y_prob_pred, axis=1)
         else:
             y_i = np.argmax(y, axis=1)
 
@@ -137,8 +144,8 @@ class SimBA(EvasionAttack):
 
             desired_label = y_i[i_sample]
 
-            current_label = np.argmax(preds, axis=1)[i_sample]
-            last_prob = preds[i_sample].reshape(-1)[desired_label]
+            current_label = np.argmax(y_prob_pred, axis=1)[i_sample]
+            last_prob = y_prob_pred[i_sample].reshape(-1)[desired_label]
 
             if self.estimator.channels_first:
                 nb_channels = x.shape[1]
--- a/art/attacks/evasion/simba.py
+++ b/art/attacks/evasion/simba.py
@@ -101,10 +101,12 @@ class SimBA(EvasionAttack):
         Generate adversarial samples and return them in an array.
 
         :param x: An array with the original inputs to be attacked.
-        :param y: An array with the original labels to be predicted.
+        :param y: An array with the true or target labels.
         :return: An array holding the adversarial examples.
         """
         x = x.astype(ART_NUMPY_DTYPE)
+        x_adv = x.copy()
+
         preds = self.estimator.predict(x, batch_size=self.batch_size)
 
         if self.estimator.nb_classes == 2 and preds.shape[1] == 1:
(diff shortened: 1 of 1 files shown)


Embeddave:
Hi @beat-buesser thank you for making this fix and inviting me to review.

I would be happy to but I'm working a short week because of a US holiday.
Would it work for you if I tested early next week?

beat-buesser:
@Embeddave Thank you very much, next week would be perfect. Happy Thanksgiving!

beat-buesser:
Hi @Embeddave What do you think about this pull request?

Embeddave:
Hi @beat-buesser was going to leave you a message today

I found a major bug in my own code in the process of testing this and needed to put out that fire first

I can check early this week

we typically run with `batch_size=1` as the two-six teams lab does when running GARD evals so I haven't needed this specific functionality

beat-buesser:
@Embeddave Thank you for the update, that should work. I was sure you had not forgotten.

Embeddave:
I have some separate feedback but I will reply on #1407 -- it's off-topic for this PR

art/attacks/evasion/simba.py
```diff
@@ -107,9 +108,15 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> n
         x = x.astype(ART_NUMPY_DTYPE)
         x_adv = x.copy()
 
-        preds = self.estimator.predict(x, batch_size=self.batch_size)
+        y_prob_pred = self.estimator.predict(x, batch_size=self.batch_size)
 
-        if self.estimator.nb_classes == 2 and preds.shape[1] == 1:
+        if not is_probability(y_prob_pred):
+            raise ValueError(
+                "This attack requires an estimator predicting probabilities. It looks like the current "
```
Embeddave, line 115: This is definitely helpful but maybe change the error message to recommend a fix?
```
                "This attack requires an estimator predicting probabilities. Try adding 
                a Softmax layer to your model and/or using the `art.BlackBoxEstimator` class"
```

beat-buesser: @Embeddave Yes, that should be helpful. I'm not sure if I understand `art.BlackBoxEstimator`?

Embeddave:
👍 
the additional commit changing the variable name and using the validator LGTM

I tested with my code that the attack is working as before when batch size = 1.
Seems to be working, AFAICT.
I would share so you can see how I'm testing, but the code involves some in-house libraries that are not public (nothing special, just not installable by anybody).

I am testing the batch size > 1 functionality now.

Embeddave (changes requested):
@beat-buesser in line ~331 (it won't let me tag in review) of `_check_params` I think you still need to remove the condition that checks whether batch size is greater than 1. I get

```
The batch size `batch_size` has to be 1 in this implementation.
```

I can test again after you remove that

beat-buesser:
@Embeddave Thank you, I have added you to the list of collaborators in the repo which upon accepting the invite in your email should allow you to make inline comments.

I think the batch size still has to be 1 only because the algorithm processes one sample at a time and does not support processing in larger mini-batches. But with this version it is now processing all samples of `x`.

art/attacks/evasion/simba.py
Embeddave: I'm not sure if I understand either 🙂 

I should have said `BlackBoxClassifier`

In truth it's not clear to me whether that would be appropriate and/or expected to use with this attack -- I thought I had read an example notebook that used the `BlackBoxClassifier` wrapper with a similar attack but now I can't find it

So, nevermind :)

but a concrete solution like "add softmax layer" I think we agree could be good?

Embeddave:
Hi again @beat-buesser I'm not sure the ability to run multiple images at a time is working as expected.

If I pass in a single `x` with multiple images, I get a different answer compared to if I pass in each image one at a time.

Here's a hopefully minimal reproducible example
```python
from functools import partial
from art.estimators.classification import PyTorchClassifier
from art.attacks.evasion import SimBA
import numpy as np
import torch
import torch.nn as nn
from torchvision import transforms, datasets
from tqdm import tqdm

def run_batch(model, x, y, normalizer=None):
    """runs a batch of images through model to get
    predictions, compute which are correct, and
    from that compute accuracy

    Parameters
    ----------
    model : torch.nn.Module
        instance of a neural network model
    x : torch.Tensor
        input to network
    y : torch.Tensor
        ground truth that predictions should match
    normalizer : robustness.Normalizer
        used to normalize inputs to network.
        Default is None, in which case no normalization is applied.

    Returns
    -------
    y_pred : list
        of class predictions, same size as y
    y_prob : list
        of float, the probabilities that the model
        assigned to the classes in ``y_pred``.
        I.e., the scalar value from softmax(output)
        indexed by argmax(softmax(output)),
        for each input in ``x``.
    acc : float
        number correct / total number of samples in batch
    """
    if normalizer is not None:
        x = normalizer(x)
    with torch.no_grad():
        out = model(x)
        y_pred = out.argmax(dim=1)
        y_prob = torch.nn.Softmax()(out)[:, y_pred].flatten()
    correct = (y_pred == y).cpu().numpy().tolist()
    acc = sum(correct) / len(correct)
    return y_pred.cpu().numpy().tolist(), y_prob.cpu().numpy().tolist(), acc

# ... set up model etc
# for simba attack, need to make model output probabilities
model.fc = nn.Sequential(
    *[model.fc, nn.Softmax(dim=1)],
)

log_or_print('setting up adversarial attacks')

# OMITTED custom code that uses `robustness` library
# to get dataloaders etc
# we're using Imagenette dataset from fastai

# constants from torchvision references training script
mean = np.asarray(constants.IMAGENET_MEAN).reshape((3, 1, 1)).astype(np.float32)
std = np.asarray(constants.IMAGENET_STD).reshape((3, 1, 1)).astype(np.float32)
    nb_classes = 1000
    input_shape = (3, 224, 224)

clf = PyTorchClassifier(
    model=model,
    clip_values=(0., 1.),
    loss=criterion,
    input_shape=input_shape,
    nb_classes=nb_classes,
    preprocessing=(mean, std)
)

attack = SimBA(classifier=clf,
               attack='dct',
               max_iter=5000,
               order='random',
               epsilon=0.9,
               freq_dim=28,
               stride=7,
               )

MAX_GOOD_BATCHES = 4

good_batches = []
n_good_batches = 0

run_batch_partial = partial(helpers.run_batch,
                            model=model,
                            normalizer=normalizer,
                            y=y)

for batch in test_loader:
    x, y = batch['img'].to(device), batch['target'].to(device)
    source_img = batch['source']
    y_true = y.cpu().numpy().tolist()

    # ---- unattacked
    log_or_print('\trunning unattacked batch')
    y_pred, correct_unat, acc_unat = run_batch_partial(x=x)

    log_or_print(
        f'\t\tbatch accuracy: {acc_unat:.3f}'
    )
    if acc_unat > 0.99:
        good_batches.append(x)
        n_good_batches += 1
        if n_good_batches > MAX_GOOD_BATCHES:
            break

x = good_batches[2]

with torch.no_grad():
    y_probs = model(normalizer(x))
y_probs = y_probs.cpu().numpy()
x_np = x.cpu().numpy()
x_at = attack.generate(x=x_np, y=y_probs)

x_at_tensor = torch.from_numpy(x_at).to(device)
y_pred_at, correct_at, acc_at = run_batch_partial(x=x_at_tensor)
log_or_print(
    f'\t\tbatch accuracy: {acc_at:.3f}'
)

# gives me batch accuracy: 0.667

for x_img in x:
    x_img = torch.unsqueeze(x_img, 0)
    with torch.no_grad():
        y_probs = model(
            normalizer(x_img)
        )
    y_probs = y_probs.cpu().numpy()
    x_img = x_img.cpu().numpy()
    x_at = attack.generate(x=x_img, y=y_probs)

    x_at_tensor = torch.from_numpy(x_at).to(device)
    y_pred_at, correct_at, acc_at = run_batch_partial(x=x_at_tensor)
    log_or_print(
        f'\t\tbatch accuracy: {acc_at:.3f}'
    )

# gives me
# batch accuracy: 0.000
# batch accuracy: 0.000
# batch accuracy: 0.000
```

Embeddave:
I can go as far as writing a whole script if you're not able to reproduce this

AFAICT `SimBa` is working *without* running multiple batches so if you don't need to add that functionality I would be fine with it the way it is, not sure about other users

I still think the other changes in this PR like the check for probabilities and the variable name change would be useful regardless

beat-buesser:
Hi @Embeddave I have created a similar script based on ART's `get_started_pytorch.py` and it achives the same accuracy with calling generate on all images or on each image separately:

```python
import torch.nn
import torch.nn as nn
import torch.nn.functional as F
import torch.optim as optim
import numpy as np

from art.estimators.classification import PyTorchClassifier
from art.utils import load_mnist


# Step 0: Define the neural network model, return logits instead of activation in forward method


class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv_1 = nn.Conv2d(in_channels=1, out_channels=4, kernel_size=5, stride=1)
        self.conv_2 = nn.Conv2d(in_channels=4, out_channels=10, kernel_size=5, stride=1)
        self.fc_1 = nn.Linear(in_features=4 * 4 * 10, out_features=100)
        self.fc_2 = nn.Linear(in_features=100, out_features=10)
        self.softmax = torch.nn.Softmax(dim=1)

    def forward(self, x):
        x = F.relu(self.conv_1(x))
        x = F.max_pool2d(x, 2, 2)
        x = F.relu(self.conv_2(x))
        x = F.max_pool2d(x, 2, 2)
        x = x.view(-1, 4 * 4 * 10)
        x = F.relu(self.fc_1(x))
        x = self.fc_2(x)
        x = self.softmax(x)
        return x


# Step 1: Load the MNIST dataset

(x_train, y_train), (x_test, y_test), min_pixel_value, max_pixel_value = load_mnist()

# Step 1a: Swap axes to PyTorch's NCHW format

x_test = x_test[0:100]
y_test = y_test[0:100]

x_train = np.transpose(x_train, (0, 3, 1, 2)).astype(np.float32)
x_test = np.transpose(x_test, (0, 3, 1, 2)).astype(np.float32)

# Step 2: Create the model

model = Net()

# Step 2a: Define the loss function and the optimizer

criterion = nn.CrossEntropyLoss()
optimizer = optim.Adam(model.parameters(), lr=0.001)
# optimizer = optim.SGD(model.parameters(), lr=0.01)

# Step 3: Create the ART classifier

classifier = PyTorchClassifier(
    model=model,
    clip_values=(min_pixel_value, max_pixel_value),
    loss=criterion,
    optimizer=optimizer,
    input_shape=(1, 28, 28),
    nb_classes=10,
)

# Step 4: Train the ART classifier

classifier.fit(x_train, y_train, batch_size=64, nb_epochs=1)

# Step 5: Evaluate the ART classifier on benign test examples

predictions = classifier.predict(x_test)
accuracy = np.sum(np.argmax(predictions, axis=1) == np.argmax(y_test, axis=1)) / len(y_test)
print("Accuracy on benign test examples: {}%".format(accuracy * 100))


from art.attacks.evasion import SimBA

attack = SimBA(classifier=classifier,
               attack='dct',
               max_iter=5000,
               order='random',
               epsilon=0.9,
               freq_dim=28,
               stride=7,
               verbose=False,
               )

# attack `all`

x_test_adv_all = attack.generate(x=x_test, y=y_test)

predictions = classifier.predict(x_test_adv_all)
accuracy = np.sum(np.argmax(predictions, axis=1) == np.argmax(y_test, axis=1)) / len(y_test)
print("Accuracy on adversarial test examples - all: {}%".format(accuracy * 100))

# attack `single`

count_correct = 0

for i in range(x_test.shape[0]):

    x_test_adv_single = attack.generate(x=x_test[[i]], y=y_test[[i]])

    prediction = classifier.predict(x_test_adv_single)
    accuracy = np.sum(np.argmax(prediction, axis=1) == np.argmax(y_test[[i]], axis=1))
    if accuracy:
        count_correct += 1

print("Accuracy on adversarial test examples - single: {}%".format(count_correct / x_test.shape[0] * 100))
```

Embeddave:
Thank you @beat-buesser for providing this script.

I will test using it with my images and models this week.

If I still get the same answer where attack success depends on the number of images passed in, I will upload a minimal reproducible example.

beat-buesser:
Hi @Embeddave Thank you for your review and testing. To prepare for the ART 1.9 release, I'll merge this PR to include the exception and the processing of mini-batches. In case there are still bugs left we can fix them with a new PR for 1.9.0 or 1.9.1.

Embeddave:
Understood, thank you @beat-buesser

PR #1612, opened 2022-03-29
Add nb_epochs and batch_size to AdversarialTrainerMadryPGD.fit

### LINKED ISSUE #1609: Inconsistent API for AdversarialTrainer
**Describe the bug**
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/defences/trainer/adversarial_trainer_madry_pgd.py#L58

The standard API for ART `fit` method is to pass in `batch_size` as a kwargs:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/estimators/pytorch.py#L107
This is what is done in the FBF AdversarialTrainer API here:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/defences/trainer/adversarial_trainer_fbf_pytorch.py#L72

However, the Madry AdversarialTrainer sets batch_size in `__init__`:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/defences/trainer/adversarial_trainer_madry_pgd.py#L58
and then uses this value internally inside its `fit` method:
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/defences/trainer/adversarial_trainer_madry_pgd.py#L105

**To Reproduce**
If you simply instantiate the Madry trainer and call `fit` with a `batch_size` kwarg, you'll get something like:
```
TypeError: new_fit() got multiple values for keyword argument 'batch_size'
```

**Expected behavior**
API should be consistent and `batch_size` should be set in `fit`, not `__init__`.

The fix is simple: move where `batch_size` is set.

beat-buesser:

# Description

This pull request adds `nb_epochs` and `batch_size` to `AdversarialTrainerMadryPGD.fit`.

Fixes #1609

## Type of change

- [ ] Bug fix (non-breaking)

### DIFF
--- a/art/defences/trainer/adversarial_trainer_madry_pgd.py
+++ b/art/defences/trainer/adversarial_trainer_madry_pgd.py
@@ -54,8 +54,8 @@ class AdversarialTrainerMadryPGD(Trainer):
     def __init__(
         self,
         classifier: "CLASSIFIER_LOSS_GRADIENTS_TYPE",
-        nb_epochs: int = 391,
-        batch_size: int = 128,
+        nb_epochs: Optional[int] = 391,
+        batch_size: Optional[int] = 128,
         eps: Union[int, float] = 8,
         eps_step: Union[int, float] = 2,
         max_iter: int = 7,
@@ -91,7 +91,13 @@ class AdversarialTrainerMadryPGD(Trainer):
         self.trainer = AdversarialTrainer(classifier, self.attack, ratio=1.0)  # type: ignore
 
     def fit(  # pylint: disable=W0221
-        self, x: np.ndarray, y: np.ndarray, validation_data: Optional[np.ndarray] = None, **kwargs
+        self,
+        x: np.ndarray,
+        y: np.ndarray,
+        validation_data: Optional[np.ndarray] = None,
+        batch_size: Optional[int] = None,
+        nb_epochs: Optional[int] = None,
+        **kwargs
     ) -> None:
         """
         Train a model adversarially. See class documentation for more information on the exact procedure.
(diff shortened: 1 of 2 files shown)


art/defences/trainer/adversarial_trainer_madry_pgd.py
```diff
@@ -54,8 +54,8 @@ class AdversarialTrainerMadryPGD(Trainer):
     def __init__(
         self,
         classifier: "CLASSIFIER_LOSS_GRADIENTS_TYPE",
-        nb_epochs: int = 391,
-        batch_size: int = 128,
+        nb_epochs: Optional[int] = 391,
```
GiulioZizzo, line 57: Although not part of the core change, is it worth adding a qualifier in the comments as to why 391 was chosen as the default argument? It's a non-standard number: is it based on the original paper? 

Should this (along with the batch size) have the Optional typing removed? If None is supplied here and then not explicitly passed in the fit it could cause a problem.

GiulioZizzo (commented):
Hi Beat, thank you for the review! I made a minor comment regarding the __init__ . Let me know what you think.

art/defences/trainer/adversarial_trainer_madry_pgd.py
beat-buesser: That's correct, if both values for `nb_epochs` are None it will now raise a ValueError with an explanation. This is to remain backward compatible and to align all trainers to the same keyword arguments.

I think the 391 should actually be 205 epochs to represent 80'000 training steps on 50'000 CIFAR10 training samples with batch_size 128. We'll update it for ART 1.11 to keep the default behaviour for ART 1.10.x

beat-buesser: I opened issue #1644 for the default value update.

PR #1778, opened 2022-07-11
Shadow models and predictions for membership attack

### LINKED ISSUE #1756: Support overlapping shadow model datasets
The current implementation of shadow model training creates disjoint train and test datasets for each shadow model: 
shadow_dataset_size = len(x) // len(self._shadow_models)
This does not enable using high numbers of shadow models, especially with small datasets.

The original Shokri et al. paper says in their experimental setup section: "There is no overlap between the datasets of the target model and those of the shadow models, but the datasets used for different shadow models can overlap with each other". 
I suggest to add an optional parameter to this class indicating whether to allow overlapping datasets (with default True).

abigailgold:
# Description
1. Enable creating shadow datasets that overlap (as opposed to only supporting disjoint datasets). At the moment only supports completely disjoint or complete overlap.
2. Enable sending the blackbox membership attack model predictions as input (instead of the model itself). This enables supporting use cases where the model or training data (X) is not available. 

Fixes #1756 

## Type of change
- [ ] Bug fix (non-breaking)

# Testing
Added test for new functionality (both for shadow models and membership attack)

**Test Configuration**:
- OS: 12.4
- Python version: 3.8

# Checklist


art/estimators/regression/pytorch.py
```diff
@@ -533,6 +533,45 @@ def compute_loss(  # type: ignore # pylint: disable=W0221
 
         return loss.detach().cpu().numpy()
 
+    def compute_loss_from_predictions(  # type: ignore # pylint: disable=W0221
```
beat-buesser, line 536: Why do we need this new method in addition to the existing method `compute_losses`?

art/estimators/estimator.py
```diff
@@ -338,6 +338,16 @@ def compute_loss(self, x: np.ndarray, y: Any, **kwargs) -> np.ndarray:
         """
         raise NotImplementedError
 
+    def compute_loss_from_predictions(self, pred: np.ndarray, y: np.ndarray, **kwargs) -> np.ndarray:
```
beat-buesser, line 341: Why do we need this new method in addition to the existing method compute_losses?

art/estimators/regression/pytorch.py
beat-buesser, line 536: Why do we need this new method in addition to the existing method `compute_losses`?

art/estimators/estimator.py
abigailgold: To serve the use case where we may not have access to training data (features) but only to the model predictions.

beat-buesser (approved):
Hi @abigailgold Thank you very much for your pull request! It looks good to me.

Issue #1799, opened 2022-07-29
random_sphere() samples points in the ball (not on the sphere)


Framartin:
**Describe the bug**
The name and documentation of the function `art.utils.random_sphere()` suggest that the function samples points on the Lp-sphere, but it samples points in the Lp-ball. 

**To Reproduce**
1. Sample N points in D dimensions using `art.utils.random_sphere`
2. Compute their p-norm

```python
import numpy as np
from art.utils import random_sphere
d=10
for p in [1, 2, np.inf]:
    points = random_sphere(nb_points=5, nb_dims=d, radius=10, norm=p)
    norm = np.linalg.norm(points, ord=p, axis=1)
    print(f'{p}-norm of {d}-dim points: {norm}')
# 1-norm of 10-dim points: [1.90252354 7.96682405 9.87868038 2.48234154 4.31938079]
# 2-norm of 10-dim points: [9.5777607  7.62736975 9.69089633 8.89324436 9.77982516]
# inf-norm of 10-dim points: [8.13604454 9.64728559 9.98291139 9.81341037 9.71205873]
```

I do agree that in high dimension, the issue becomes negligible for the 2 and inf norms, but it is not for the 1-norm.
```python
# 1-norm of 10000-dim points: [8.84152977 5.19254224 7.03462032 1.38944847 6.84087878]
# 2-norm of 10000-dim points: [9.99832332 9.99853434 9.99970224 9.99974475 9.99949171]
# inf-norm of 10000-dim points: [9.99979021 9.99997746 9.99857432 9.99995599 9.99925035]
```

**Expected behavior**
I expect the p-norm of all points to be equal to `radius`. 

**System information (please complete the following information):**
 - OS: any
 - Python version: 3.8.8
 - ART version or commit number: 1.11.0

beat-buesser:
Hi @Framartin Thank you very much for raising this issue and the code to reproduce your results!

@nmegiddo What do you think?

nmegiddo:
I am not sure what the intended use of this function is, but it clearly produces points in the ball rather than on the sphere. 
The question is what to do: Change the name of the function to random_ball?  Change the code so it produces points on the sphere?  Both can creates problems.

Further comments:
1. The inf-norm is the most transparent case:  
      res = np.random.uniform(-radius, radius, (nb_points, nb_dims))
[...]

PR #1800, opened 2022-07-30
Fix random_sphere() with norm=1

Framartin:
# Description

Partial fix of #1799 :
- Fix `art.utils.random_sphere()` when `norm=1` which now correctly sample uniformly in the L1 ball. The sampling method comes from [this reference](https://mathoverflow.net/a/9188/488767), which is also used by [foolbox](https://github.com/bethgelab/foolbox/blob/12abe74e2f1ec79edb759454458ad8dd9ce84939/foolbox/attacks/gradient_descent_base.py#L197).
- Add two unit tests to cover the issue
- The docstring of `art.utils.random_sphere()` now specifies that the function samples in the p-norm ball (instead of the surface of the sphere)

## Type of change

- [x] Bug fix (non-breaking)

# Testing

The norm in low and high dimensions is close to what was expected.

```python
import numpy as np
from art.utils import random_sphere
p=1
for d in [10, 10000]:
    points = random_sphere(nb_points=5, nb_dims=d, radius=10, norm=p)
    norm = np.linalg.norm(points, ord=p, axis=1)
    print(f'{p}-norm of {d}-dim points: {norm}')

# 1-norm of 10-dim points: [9.42352505 8.55454679 8.58168686 9.40038438 9.09817243]
# 1-norm of 10000-dim points: [9.99686352 9.99781933 9.99959791 9.99949611 9.99997998]
```

Check  2D scatterplot

```python
import matplotlib.pyplot as plt
points = random_sphere(nb_points=10000, nb_dims=2, radius=1, norm=1)
plt.scatter(points[:,0], points[:,1])
```
scaltter_l1_ball

The two unit tests should not be flaky, as the largest norm over 100000 runs is always approximately 1 order of magnitude lower than the chosen threshold:
```python
# max diffrence between the norm of 100K points in the L1 unit balls and 1.
a = [np.abs(np.sum(np.abs(random_sphere(1, 10000, 1, 1)), axis=1) - 1.0) for i in range(100000)]
np.max(a) < 1e-2  # True
print(np.max(a)) # 0.0011441126974902627
# same for L2
a = [np.abs(np.linalg.norm(random_sphere(1, 10000, 1, 2), axis=1) - 1.0) for i in range(100000)]
np.max(a) < 1e-2  # True
print(np.max(a)) # 0.001347197437172154
# same for Linf
a = [np.abs(np.max(np.abs(random_sphere(1, 10000, 1, np.inf)), axis=1) - 1.0) for i in range(100000)]
np.max(a) < 1e-2  # True
print(np.max(a)) # 0.0011067893448932775
```

**Test Configuration**:

- OS: MacOS
- Python version: 3.8.8
- ART version or commit number: e7f3b7faa68f5efac082ed2b2b79b8ee3dda78a5

# Checklist


Framartin:
My apologies, I messed up my git tree (and could not rebase easily to sign off my previous commits). #1802 replaces this PR.

PR #1956, opened 2022-12-08
DP-InstaHide Trainer Implementation

f4str:
# Description

Implementation of the DP-InstaHide trainer. This trainer builds upon the previously implemented data augmentation defenses. Also includes a jupyter notebook demo showing how to use the trainer and its effectiveness on MNIST (a reference was added to the notebook README).

As a consequence, the following two bugs/issues were resolved
* Call the `check_and_transform_label_format` function when calling the `compute_loss` function in the `TensorFlowV2Classifier` and `KerasClassifier` classes. This fixes the argmax issue when passing integer labels. (Consolidated from PR #1960)
* Create a new property `_probability_labels` in the `PyTorchClassifier` and have it set to True when using `nn.CrossEntropyLoss`. This allows the `reduce_labels()` function to check if the labels are one-hot encoded or a probability distribution and handle it accordingly. This allows support for probability distribution labels. (Consolidated from PR #1967)

Fixes #1827 
Fixes #1959 
Fixes #1966

## Type of change

Please check all relevant options.

- [x] Improvement (non-breaking)
- [x] Bug fix (non-breaking)
- [x] New feature (non-breaking)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# Testing

Please describe the tests that you ran to verify your changes. Consider listing any relevant details of your test configuration.

- [x] Test the DP-InstaHide trainer in PyTorch
- [x] Test the DP-InstaHide trainer in TensorFlow
- [x] Test the DP-InstaHide trainer in Keras

**Test Configuration**:
- OS
- Python version
- ART version or commit number
- TensorFlow / Keras / PyTorch / MXNet version

# Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1967, opened 2022-12-13
Add Support for Probability Distribution Labels in PyTorchClassifier

### LINKED ISSUE #1966: Support Probability Distribution Labels in PyTorchClassifier
PyTorch's `nn.CrossEntropyLoss` function supports probability distribution labels. However, since the `PyTorchClassifier` always sets `_reduce_labels = True` when using this loss function, it is not possible to use probability distribution labels. The solution is to never reduce labels for this loss function as the result is the same when using vector labels or one-hot encoded labels.

When calling `classifier.fit()` or `classifier.compute_loss()` using `nn.CrossEntropyLoss`, labels passed as a probability distribution should be kept as is and not reduced. No clear solution has been decided yet, instead ideas are below.

Create a new `_probability_labels` property in the `PyTorchClassifier` and have it set to True when using `nn.CrossEntropyLoss`. This will allow to check in the `reduce_labels()` function if the provided labels are one-hot encoded or probability distribution labels. However, this would introduce additional computation time to check the labels and may also break some of the tests.

Manually checking the labels beforehand to determine whether to reduce labels. But this would introduce extra complexity and everything should be left inside the `reduce_labels()` function.

Setting the `_reduce_labels` property to False in the `PyTorchClassifier` when using `nn.CrossEntropyLoss` so labels are always reduced is also an option since the results are the same. However, this breaks some of the unit tests for the PyTorch randomized smoothing implementation.

Use the `kwargs` in the `.fit()` and `.compute_loss()` functions to pass parameters to override the current `_reduce_labels` and `_int_labels` fields. This will preserve compatibility, but is not the most elegant solution.

**Additional context**
N/A

f4str:
# Description

Create a new property `_probability_labels` in the `PyTorchClassifier` and have it set to True when using `nn.CrossEntropyLoss`. This allows the `reduce_labels()` function to check if the labels are one-hot encoded or a probability distribution and handle it accordingly. This allows support for probability distribution labels.

Additionally do the following minor changes: 
* Clean up some of the code/formatting for the label checking in `PyTorchClassifier`
* Add `BCEWithLogitsLoss` as one of the loss functions to be checked since the same reduction should occur as with `BCELoss`
* Removed commented code for label reduction in the PyTorch regressor, mainly to avoid false positives when searching the usage of the `reduce_labels()` function

Fixes #1966 

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

- [ ] Test A
- [ ] Test B

**Test Configuration**:
- OS
- Python version

# Checklist

### DIFF
--- a/art/estimators/classification/pytorch.py
+++ b/art/estimators/classification/pytorch.py
@@ -272,7 +272,7 @@ class PyTorchClassifier(ClassGradientsMixin, ClassifierMixin, PyTorchEstimator):
         # Check if the loss function supports probability labels and probability labels are provided
         if self._probability_labels and len(y.shape) == 2:
             if isinstance(y, torch.Tensor):
-                is_one_hot = torch.equal(y, y.int())
+                is_one_hot = torch.equal(y.floor(), y.ceil())
             else:
                 is_one_hot = np.array_equal(y, y.astype(np.int32))
             if not is_one_hot:  # probability labels
--- a/art/estimators/classification/pytorch.py
+++ b/art/estimators/classification/pytorch.py
@@ -153,7 +153,7 @@ class PyTorchClassifier(ClassGradientsMixin, ClassifierMixin, PyTorchEstimator):
             self._reduce_labels = True
             self._int_labels = True
             self._probability_labels = True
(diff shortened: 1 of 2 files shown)


f4str:
Consolidated to PR #1956 since this change can be properly tested there.

PR #2124, opened 2023-04-24
Implement Default Training Loop for `TensorFlowV2Classifier`

### LINKED ISSUE #2083: Default `train_step` for `TensorFlowV2Classifier` and Variations
The `TensorFlowV2Classifier` requires the `train_step` parameter when initialized for training. However, most of the time, this `train_step` is always the same:

```python
def train_step(model, images, labels):
    with tf.GradientTape() as tape:
        predictions = model(images, training=True)
        loss = loss_object(labels, predictions)
    gradients = tape.gradient(loss, model.trainable_variables)
    optimizer.apply_gradients(zip(gradients, model.trainable_variables))
```

Rather than requiring the user to always provide the same `train_step` this default one should be used. In this case, the user will be able to either use their own custom `train_step` or use the default one provided.

If the user does not provide a `train_step` function when initializing the `TensorFlowV2Classifier` use the default one above. The same should be done for all implementations of the TensorFlowV2 classes such as `TensorFlowV2RandomizedSmoothing`, `TensorFlowV2DeRandomizedSmoothing`, etc.

Rather than using this `train_step` function, the class can follow a similar structure as the `PyTorchClassifier` where it takes a classifier and performs the training. However, this would break backwards compatibility so it is not the best option.

**Additional context**
The default `train_step` function will have the `@tf.function` decorator so it will be optimized by the TensorFlow computational graph. This will greatly speed up the training process when using the default `train_step`. This decorator should also be documented so the users are aware this can be used to speed up training.

f4str:
# Description

A new `optimizer` parameter was added to the `TensorFlowV2Classifier` and its derivatives: `TensorFlowV2RandomizedSmoothing` and `TensorFlowV2RandomizedSmoothing`. This new parameter is now used in the `.fit(...)` methods to create a default `train_step` training loop that is optimized with the `@tf.function` decorator. The existing `train_step` parameter can still be provided to override the default training loop. This preserves backwards compatibility and ensures everything still works. The docstrings have been updated accordingly.

All test cases and relevant examples/notebooks have been updated to now use the default (optimized) training loop rather than creating one manually. This simplifies the setup for the TensorFlow v2 classifiers and makes it more similar to other estimators like the `KerasClassifier` and `PyTorchClassifier`. Note that all test cases still work the same as before and did not need to be modified at all, but this change just cleans things up.

By using this optimized default training loop, it will significantly improve training performance for the `TensorFlowV2Classifier` and its derivatives.

Fixes #2083 

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

- [x] Existing test cases changed to use the new default training loop

**Test Configuration**:
- OS
- Python version

# Checklist


art/attacks/evasion/brendel_bethge.py
```diff
@@ -2055,6 +2055,7 @@ def logits_difference(y_true, y_pred):
                 nb_classes=estimator.nb_classes,
                 input_shape=estimator.input_shape,
                 loss_object=self._loss_object,
+                optimizer=estimator._optimizer,
```
beat-buesser, line 2058: Should we use the new property:
```suggestion
                optimizer=estimator.optimizer,
```

beat-buesser, line 2058: Should we use the new property:
```suggestion
                optimizer=estimator._optimizer,
```

art/attacks/evasion/auto_projected_gradient_descent.py
```diff
@@ -203,6 +203,7 @@ def __call__(self, y_true: tf.Tensor, y_pred: tf.Tensor, *args, **kwargs) -> tf.
                     nb_classes=estimator.nb_classes,
                     input_shape=estimator.input_shape,
                     loss_object=_loss_object_tf,
+                    optimizer=estimator._optimizer,
```
beat-buesser, line 206: Should we use the new property:
```suggestion
                    optimizer=estimator.optimizer,
```

art/attacks/evasion/brendel_bethge.py
beat-buesser, line 2058: Should we use the new property:
```suggestion
                optimizer=estimator.optimizer,
```

art/attacks/evasion/auto_conjugate_gradient.py
```diff
@@ -224,6 +224,7 @@ def __call__(self, y_true: tf.Tensor, y_pred: tf.Tensor, *args, **kwargs) -> tf.
                     nb_classes=estimator.nb_classes,
                     input_shape=estimator.input_shape,
                     loss_object=_loss_object_tf,
+                    optimizer=estimator._optimizer,
```
beat-buesser, line 227: Should we use the new property:
```suggestion
                    optimizer=estimator.optimizer,
```

beat-buesser (changes requested):
Hi @f4str Thank you very much for implementing a default training step for TensorFlow v2 classifiers! I have added a few suggestions, what do you think?

f4str:
Hi @beat-buesser thank you for the comments! For consistency, `optimizer` currently uses the private attribute instead of the public since `loss_object` and `train_step` use the private attribute. So all of these should use either the private attribute or public property to stay consistent.

Since ART is shifting from private attributes to public properties, I think it may be worth changing all of these (`optimizer`, `loss_obejct`, and `train_step`) to use the public property, but only for the affected files in this PR alone. Any thoughts on this?

beat-buesser (approved):

beat-buesser:
Hi @f4str Thank you very much, the updates look good to me.

Issue #1943, opened 2022-11-30
Code modifications in "pytorch_object_detector.py"


ShengYun-Peng:
**Is your feature request related to a problem? Please describe.**
In eval6, Armory is using ["pytorch_object_detector.py"](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/main/art/estimators/object_detection/pytorch_object_detector.py) for the multi-object tracking scenario. Currently, a single input image is passed into the preprocessor under a [for loop](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/c226388533f0029d03f2c40bdd9a4e7b853e5f30/art/estimators/object_detection/pytorch_object_detector.py#L209). However, our tracking defense relies on preprocessing multiple frames at one time, we have to move that chunk of code out of the for loop in "pytorch_object_detector.py" to make the [preprocessor](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/c226388533f0029d03f2c40bdd9a4e7b853e5f30/art/estimators/object_detection/pytorch_object_detector.py#L223) accept full video frames. 

**Describe the solution you'd like**
Replace 
https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/c226388533f0029d03f2c40bdd9a4e7b853e5f30/art/estimators/object_detection/pytorch_object_detector.py#L209-L228 

with
```python
if isinstance(x, np.ndarray):
    if self.clip_values is not None:
        x_grad = transform(x / self.clip_values[1]).to(self.device)
    else:
        x_grad = transform(x).to(self.device)
    x_grad.requires_grad = True
else:
    x_grad = x.to(self.device)
    if x_grad.shape[3] < x_grad.shape[1] and x_grad.shape[3] < x_grad.shape[2]:
        x_grad = torch.permute(x_grad, (0, 3, 1, 2))

image_tensor_list_grad.append(x_grad)
x_grad_1 = x_grad
x_preprocessed_i, y_preprocessed_i = self._apply_preprocessing(
    x_grad_1, y=y_tensor, fit=False, no_grad=False
)
for i_preprocessed in range(x_preprocessed_i.shape[0]):
    inputs_t.append(x_preprocessed_i[i_preprocessed])
    y_preprocessed.append(y_preprocessed_i[i_preprocessed])
```

**Describe alternatives you've considered**
[...]

PR #2180, opened 2023-06-06
Optimize PyTorch Classifiers and Object Detectors

### LINKED ISSUE #1637: Update argument `channels_first` in `PyTorchObjectDetector` and related estimators
The argument `channels_first` in `PyTorchObjectDetector` is currently unused. We should either remove it or  set the correct value expected by the supported models.

### LINKED ISSUE #2157: Optimize `fit` and `predict` loops for PyTorch Estimators
All PyTorch estimators currently index slice numpy arrays in their loops for the `fit` and `predict` methods. Some of these will even put the data on GPU prior to batching which uses an unnecessary amount of GPU VRAM. A more optimized approach is to use the PyTorch dataloader and only move the current batch in the GPU.

Modify the `fit` and `predict` methods for all PyTorch estimators to use dataloaders and only move the current batch to the GPU

This will not only optimize the 

N/A

**Additional context**
The same should be done for all TensorFlowV2 estimators. Some estimators are already using the TensorFlow dataset, but others are still doing numpy index slicing. This can either be done in the same PR or a separate PR.

### LINKED ISSUE #2173: Remove Extraneous Loop for Loss Gradients Calculation in `PyTorchObjectDetector`
Due to the issue reported in <https://github.com/pytorch/vision/issues/4153>, the current implementation of the `PyTorchObjectDetector` loops through each input to calculate the gradients. However, this issue has been resolved in `torch>=1.10` and `torchvision>=0.11` which no longer makes this loop necessary.

The `loss_gradient` method in the `PyTorchObjectDetector` should be simplified to no longer need to loop through each individual input and use the entire batch similar to what is done in the `PyTorchYolo` estimator. Since the estimator already raises an exception when `torch==[1.8, 1.9]` and `torchvision==[0.9, 0.10]` is used, no other version checks are needed since the bug has been fixed in `torch>=1.10` and `torchvision>=0.11`.

This will most likely change the gradient values as the entire batch is being processed at once, so the unit tests will need to be updated as well.

N/A

**Additional context**
N/A

f4str:
# Description

Use PyTorch datasets and dataloaders to optimize the `fit` and `predict` methods for the following PyTorch estimators: `PyTorchClassifier`, `PyTorchRegressor`, `PyTorchRandomizedSmoothing`, `PyTorchObjectDetector`, `PyTorchYolo`. This not only speeds up the batching process, but also will only move the current batch to the GPU rather than the entire dataset. This will significantly reduce the VRAM usage.

~Optimize the `predict` method for the `TensorFlowV2Classifier` by using a TensorFlow dataset and the `@tf.function` decorator to speed up inference.~

Significant changes to the `PyTorchObjectDetector` estimator:
* The `channels_first` property is now properly used to transpose inputs accordingly.
* Unify the input preprocessing in a similar way that the `PyTorchYolo` estimator does. This simplifies the implementation and reduces redundant code.
* Remove the extraneous loop when computing the loss gradients. The underlying bug has been fixed as of `torch>=1.10` so the entire batch can be processed at once now. Due to batch processing rather an one sample at a time, the loss gradients have now changed. The unit tests have been adjusted accordingly.

Also removed the PyTorch version checking from the `PyTorchYolo` estimator as the bug only applies to the torchvision models and does not affect the external YOLO models.

Fixes #2157 
Fixes #1637 
Fixes #2173 

## Type of change

- [ ] Bug fix (non-breaking)

# Testing

- [x] Unit tests for PyTorch and TensorFlow estimators work as expected
- [x] Modified unit tests for `PyTorchObjectDetector`
- [x] Modified unit tests for `PyTorchFasterRCNN`

**Test Configuration**:
- OS
- Python version

# Checklist


f4str:
This may also potentially resolve #1943 but other issues may still persist.

art/estimators/object_detection/utils.py
```diff
@@ -88,3 +92,50 @@ def convert_pt_to_tf(y: List[Dict[str, np.ndarray]], height: int, width: int) ->
...
+
+    # Convert images into tensor
+    if isinstance(x, np.ndarray):
+        x_tensor = torch.from_numpy(x)
+    else:
+        x_tensor = x
+
+    # Convert labels into tensor
+    if y is not None and isinstance(y, list) and isinstance(y[0]["boxes"], np.ndarray):
+        y_tensor = []
```
beat-buesser, line 122: Would we be able to remove the "type: ignore" in line 139 if we add more specific typing here in line 122 like
```suggestion
        y_tensor: List[Dict[str, "torch.Tensor"]] = []
```
If yes, should we move this initialization before this if-block?

beat-buesser (approved):
Hi @f4str Thank you very much for these improvements to the estimators! They will be very useful for many users of ART. I have one suggestion, what do you think?

art/estimators/object_detection/utils.py
f4str: The `mypy` issue is actually due to the `None` type that `y` can be. The `type: ignore` was actually here before, I've just abstracted this segment of code out (since it is now reused) and made it a function. 

Attempting to make the typing changes to correct this needs to go a couple levels upwards which can makes thing overcomplicated.

f4str:
Hi @beat-buesser thank you for approving the PR. I have reverted the changes to the `TensorFlowV2Classifier` as the `tests/attacks/evasion/test_sign_opt.py` test case runs for over 4 hours with this optimization. Since this is difficult to debug, it is simpler to omit this change for now (since the main focus of this PR is the PyTorch classifiers) and revisit it later.

I've also responded to your comment about the mypy issue.

art/estimators/object_detection/utils.py
f4str: This should now be resolved

beat-buesser:
@f4str Thank you very much!

PR #2263, opened 2023-08-30
Remove in-place operations

### LINKED ISSUE #2262: `PyTorchYolo` and `PyTorchObjectDetector` modify the original numpy array
**Describe the bug**
The `PyTorchYolo` and `PyTorchObjectDetector`object detection estimators modify the original numpy array. This occurs because the `torch.from_numpy` and in-place torch methods are being used which will re-use the memory from the original numpy array.

**To Reproduce**
Create a `PyTorchYolo` model with the `clip_values` set to `(0, 255)`. Create any numpy image and pass it to the model via the `predict` method. The original numpy array will be modified.

The following snippets show a simplified version of what is happening:

```python
x = np.array([1])
y = torch.from_numpy(x)
y += 1

print(x)  # [2]
print(y)  # [2]
```

```python
x = np.array([1])
y = torch.from_numpy(x)
y = y + 1

print(x)  # [1]
print(y)  # [2]
```

```python
x = np.array([1])
y = torch.tensor(x)
y += 1

print(x)  # [1]
print(y)  # [2]
```

**Expected behavior**
The original numpy array should not be modified. This can be fixed by either making a copy or not using in-place operations. One unknown point is how much of the inputs should be copied (i.e., only images or both images and labels).

**Screenshots**
N/A

 - OS
 - Python version

beat-buesser:
# Description

This pull request removes in-place operations that overwrite the input data.

Fixes #2262

## Type of change

- [x] Bug fix (non-breaking)

### DIFF
--- a/art/estimators/object_detection/pytorch_yolo.py
+++ b/art/estimators/object_detection/pytorch_yolo.py
@@ -103,8 +103,8 @@ def translate_labels_x1y1x2y2_to_xcycwh(
         labels[:, 2:6] = label_dict["boxes"]
 
         # normalize bounding boxes to [0, 1]
-        labels[:, 2:6:2] /= width
-        labels[:, 3:6:2] /= height
+        labels[:, 2:6:2] = labels[:, 2:6:2] / width
+        labels[:, 3:6:2] = labels[:, 3:6:2] / height
 
         # convert from x1y1x2y2 to xcycwh
         labels[:, 4] -= labels[:, 2]
@@ -290,7 +290,7 @@ class PyTorchYolo(ObjectDetectorMixin, PyTorchEstimator):
 
             if not self.channels_first:
                 x_tensor = torch.permute(x_tensor, (0, 3, 1, 2))
-            x_tensor /= norm_factor
+            x_tensor = x_tensor / norm_factor
 
             # Set gradients
             if not no_grad:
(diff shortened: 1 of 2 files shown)


art/estimators/object_detection/pytorch_yolo.py
```diff
@@ -103,8 +103,8 @@ def translate_labels_x1y1x2y2_to_xcycwh(
         labels[:, 2:6] = label_dict["boxes"]
 
         # normalize bounding boxes to [0, 1]
-        labels[:, 2:6:2] /= width
-        labels[:, 3:6:2] /= height
+        labels[:, 2:6:2] = labels[:, 2:6:2] / width
+        labels[:, 3:6:2] = labels[:, 3:6:2] / height
```
f4str, line 107: This isn't necessary since `labels` is a newly created tensor. But the functionality is still the same.

f4str (approved):
Hi @beat-buesser thank you fixing this issue! I've tested it with the `attack_adversarial_patch_pytorch_yolo.ipynb` demo notebook and it is working once again without issues.

One minor comment that the change in `translate_predictions_xcycwh_to_x1y1x2y2` is not needed, but functionality-wise its the same. Feel free to keep the change or revert it back.

PR #2382, opened 2024-01-09
[FIX + Enhancement] `FGM` and `PGD`: fix L1 and extend to Lp

### LINKED ISSUE #2381: L¹ `FGM` is wrong + extend to all p >= 1
Hello,

I'm not sure, but I think the `FGM` extension to $L¹$ norm is not correct.

From what I can read [here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/blob/044f87e22f872e430b21ed95637577b192b6c8b9/art/attacks/evasion/fast_gradient.py#L441), it seems to me that the current version implements (essentially)
$$\text{noise direction}=\frac{\nabla}{\Vert\nabla\Vert_1}$$
when
$$\text{noise direction}=(0, \dots, 0, \text{sign}(a_i), 0, \dots, 0),\quad(i=\text{argmax}_j\vert\nabla_j\vert)$$
gives a higher inner product $\langle\nabla,\text{noise direction}\rangle$ for the same $L¹$ budget.

Indeed, in both cases $\Vert\text{noise direction}\Vert_1=1$ while the first and second options respectively give $\Vert\nabla\Vert_2/\Vert\nabla\Vert_1$ and $\Vert\nabla\Vert_{\infty}$. The latter is of course bigger due to Hölder's inequality.

------
_Edit:_ See [here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/issues/2381#issuecomment-1883154689) for generalization to all $p\in[1, +\infty]$.

eliegoudout:
# Summary for review

## Object of the PR

It corrects the $L^1$ extension of FGM evasion attack (+ PGD) and properly extends them to $L^p$ for any $p\geqslant 1$. The conversation started [here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/issues/2381), but essentially, I fix the current extension of FGM and PGD to $L^1$ and propose a proper extension to $L^p, (p\geqslant 1)$, where the added noise is defined by
$$\text{noise direction}:=\left(\frac{\vert\nabla\vert}{\Vert\nabla\Vert_q}\right)^{q/p}\text{sign}(\nabla)$$
with $\frac{1}{p}+\frac{1}{q} = 1$ (and some abuse of notation when $p=1$ or $p=+\infty$). This is optimal since it maximizes $\langle\nabla,\text{noise direction}\rangle$ subject to the constraint $\Vert\text{noise direction}\Vert_p = 1$.

Previously, for $p=1$, the package used $\text{noise direction} = \frac{\vert\nabla\vert}{\Vert\nabla\Vert_1}$ instead of the better and optimal $\text{noise direction}=(0,…,0,\text{sign}(\nabla_i),0,…,0),(i=\text{argmax}_j|\nabla_j|)$.

## What's changed?

1. `FastGradientMethod` and `ProjectedGradientDescent` now allow `norm` to be any $p\geqslant 1$, instead of only $1, 2$ or $+\infty$,
2. In both cases, the $p=1$ case is corrected and should provide better results (if anyone cares to make a comparison, that would be nice),
3. Since PGD resorts to projection to ensure that $\Vert noise\Vert_p\leqslant\varepsilon$, the related `projection`/`_projection` functions had to be updated:
    3.a. They now work for any $p\geqslant 1$ (in fact $p>0$),
    3.b. When $p\neq+\infty$, the projection is simply made by rescaling the vector, which is suboptimal when $p\neq 2$. This behaviour was already here when $p=1$.
    3.c. The supboptimal behaviour is explicited via the addition of the keyword `suboptimal=True` (by default) and in the `__doc__`.
    3.d. An "optimal" projection is something more computationally intensive to find (via an iterative method) and was not added in this PR (future work?)
    3.e. An optimal implementation was already in the files for $p=1$ so I plugged it inside `projection` from `art.utils.py`, but it remains inaccessible from the attack classes (no control over the `suboptimal` kw).
    3.f. It is now possible to pass a different radius of projection sample-wise (see [doc](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382#issuecomment-1913872823) for the way to do it),
    3.g. For a given sample, the special case of imposing feature-wise bound when $p=+\infty$ was already implemented, _i.e._ $\vert x_i\vert\leqslant\varepsilon_i$ for every feature $i$. This can be generalized to any $L^p$ norm working with the condition $\Vert x\Vert_{p, 1/\varepsilon}\leqslant 1$ where $\Vert x\Vert_{p, w}:=\left(\sum_{i}w_i\vert x_i\vert^p\right)^{1/p}$. This extentions is left for future work also.
4. I added a unit test for FGM with $p=10$.
5. _One point of attention:_ When $p=1$, the implementation I used for $\text{noise direction}$ in numpy and pytorch finds the ghighest coefficient in the gradients, and turns it into $\pm 1$ while putting all the others to $0$. I had trouble doing this in tensorflow, so I used an approach I believe is slower, which, in the end, splits the $\pm 1$ across all highest coefficients (in absolute value). As a result, when several coefficients of the gradients are maximal in absolute value, the tensorflow implementation gives a different noise direction (but with the same value of $\langle\nabla,\text{noise direction}\rangle$) than numpy of pytorch implementations. I don't know if this is a problem or not.
6. [Found](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382#discussion_r1581735593) and hotfixed ([7eb30c4](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382/commits/7eb30c4978b61527bdedac982092d82f6902e34d)) Momentum Iterative Method computation bug, for NumPy and PyTorch frameworks. Tensorflow disabled ([5c74f1e](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382/commits/5c74f1eeedad17f7243942db8a10ff91b36dc489)).

## Please give your feedbacks

I will be glad to hear what you think. Specifically, I'm not at all an expert in tensorflow and pytorch, so I'm not really sure about the potential type cast issues. I know I had a bit of trouble debugging the code in the first place.

Also, I think there may be potential for improved documentation or unit tests, but I would like to hear what you have to say about this.

All the best!
Élie

----------------
## Archive
Conversation started [here](https://github.com/Trusted-AI/adversarial-robustness-toolbox/issues/2381), noticing previous `FGM` extension to $L^1$ norm was not optimal and that other $L^p$ spaces were not supported.

Tasks:
- [x] identify changes to make
- [x] Implement FGM changes
- [x] Extend projection from `art.utils.py`. For this:
    - [x] Add `suboptimal` keyword to allow a first approach for any $p > 0$. Default value is `True`, which corresponds to current implementation
    - [x] Test implemented `projection_l1_1`and `projection_l1_2` and incorporate them when `suboptimal=False` for $p=1$.
    - [x] I renamed `test_projection` to `test_projection_norm` because it's the only thing it does.
    - Cases $0 < p < 1$ and $1 < p < +\infty$ remain `NotImplemented` when `suboptimal=False` for now.
- [x] Pass FGM tests
- [x] Implement PGD changes
    - [x] `_compute_perturbation` changes: $L^1$ fix + general $p$ extension
    - [x] `_projection` changes: Follow `art.utils.projection` mods
- [x] Pass PGD tests
- [x] Fix Momentum Iterative Method: wrong `momentum` update.

---------

_Edit:_ As I [pointed out](https://github.com/Trusted-AI/adversarial-robustness-toolbox/issues/2381#issuecomment-1883154689), it would be better (imo) to entirely generalize to any $p\in[1, +\infty]$ with
$$\text{noise direction}:=\left(\frac{\vert\nabla\vert}{\Vert\nabla\Vert_q}\right)^{q/p}\text{sign}(\nabla)$$
and the appropriate abuses of notations, which is more natural.

---------

Closes #2381.


eliegoudout:
I already exended ` art/attacks/evasion/fast_gradient.py`, and changed the $L^1$ test accordingly. But when trying to add another ($p=10$) test, I noticed that `FGM` uses a `projection` function.

Can anyone tell me why that is? There should be no projection in vanilla `FGM`. I would understand for `PGD`, but it already makes the `FGM` tests fail.

Thanks!

beat-buesser:
Hi @eliegoudout I think for FGSM the projection to `eps` should not have any effect if the perturbation is calcualted correctly to be of size `eps_step`. For PGD the projection will make sure the perturbation stays within `eps`.

eliegoudout:
I agree that projection should have any impact but the function is still called, which makes tests fail as long as `projection` is not extended.

I thought about extending projection, and as it happens, it's a non trivial problem, and the only open source implementation [I found](https://github.com/Optimizater/Lp-ball-Projection) was stale for 2 years, but might still be very good. I have not had time to try it yet, but it's on todo list.

All of a sudden, it will sadly add a bit of complexity to the projection part and to ART. Is that a problem?

eliegoudout:
Hi,
- `FGM` is now fully and properly extended to $1\leqslant p\leqslant +\infty$.
- Related tests pass (and I added one for $p=10$ for future reference.
- `projection` now supports all $p > 0$ in `suboptimal` mode (_i.e._ previous mode anyway, enabled by default with keyword `suboptimal`)
- It has yet to be implemented for $0 < p < 1$ or $1 < p < +\infty$
- _The implementation for this which I linked above seems to be **completely off**_. Maybe I broke something, but it doesn't work at all from what I could try.
- I updated tasks list.
- I think it is worth mentioning that in current release, not only PGD for $L^1$ is based on a wrong noise update (which I now fix), but also, it uses a simple rescaling for $L^1$ projection, which is also suboptimal even though `projection_l1_1/2` are already implemented. Since the projection is not a trivial problem, either it turns out that the computation is fast enough to be negligible, or I think it would be a good idea to add the `suboptimal` (or another word) option to `PDG`.

Cheers!

beat-buesser:
Hi @eliegoudout Thank you very much! Adding a bit of complexity is ok if it leads to an accurate calculation. I'll take a closer look soon. I like the `suboptimal` argument in `projection` Do you think this PR is ready for review (it's currently in draft)?

eliegoudout:
Thanks for your feedback! If we restrict this PR to `FGM` then it is almost ready for review (I would need to remove some "TO DO" I wrote in `PGD` files. On the other hand, if we want to merge PGD in the same PR, I think I might get it ready early next week (or least likely in the end of this week).

Adding good projection for every p would still require some more work and can be left for even later.

eliegoudout:
Trying to pass the final tests, I now understand (I think) that the `ndarray` option for `eps` in `projection` is to provide feature-wise bounds, not sample-wise bounds as I originally thought. This was not documented I think so i will try to reconcile both possibilities (pixel-wise and sample-wise) and document the behaviour.

eliegoudout:
Regarding my previous message I think I need some opinion to go further.

_**The problem:**_
Currently `projection` takes `eps`  either as a scalar or a `ndarray`, but the behaviour in undocumented.

_**What I think should be good for the user:**_
If they want,
- to be able to use sample-wise `eps` (different eps for each sample) but also,
- feature-wise `eps`, which corresponds to projecting onto the unit weighted $L^p$ ball, with weighted norm defined as
$$\Vert x\Vert_{p, \varepsilon} = \big(\sum_{i}\frac{1}{\varepsilon_i}\vert x_i\vert^p\big)^{1/p}.$$

_**Current implementation**_
Allows both but prioritizes feature-wise, because `eps` is first broadcasted to `values.shape`. For example, to set sample-wise epsilons for mnist images, the user has to input `eps.shape = (n_samples, 1, 1, 1)` or `eps.shape = (n_samples, 28, 28, 1)` for example, which seems weird to me.

_**What I think looks better:**_
I propose the following behaviour, relying on the fact that `values.ndim >= 2` since samples are always at least one dimensional (never just scalars):
> :param eps: If a scalar or a 1-D array, the sample-wise L_p norm of the ball onto which the projection occurs.
                Otherwise, must have the same `ndim` as `values`, in which case the projection occurs onto the weighted
                unit L_p ball, with weight `1 / eps` (currently supported only with infinity norm).

As you can see, this implies that when using feature-wise bounds, the user **must** provide `eps` with same `eps.ndim` as `values.ndim`.
This is necessary to allow both sample-wise and feature-wise bounds, because in the case where `values.shape = (10, 10)` for example (10 samples which are 1-D arrays of length 10), how would `projection` interpret `eps.shape = (10,)`?
In a way, this is prioritizing sample-wise bounds.

_**Possible Solutions:**_
- The implemented one (looks very weird to me)
- What I propose above. This seems to be a breaking change, as test `test_9a_keras_mnist` seems to generate an `eps.shape = (28, 28, 1)` with `values.shape = (10, 28, 28, 1)`,
- Add another kw option such as `samplewise` which would default to `False`. If `False`, current behaviour is maintained, if `True`, then 1d arrays are treated as in the solution I propose. I think this kw solution is heavy and quite bad tbh.

Do you have any opinion? Is the solution a propose really breaking anything?

Thanks!

eliegoudout:
In the end, I decided to revert to previous implementation, which prioritizes the feature-wise bounds, since it doesn't break anything and it may make more sense anyways, for attacks designed for batch generation.

I documened as follows the use of `eps` in `projection`/`_projection`:
> :param eps: If a scalar, the norm of the L_p ball onto which samples are projected. Equivalently in general,
            can be any array of non-negatives broadcastable with `values`, and the projection occurs onto the
            unit ball for the weighted L_{p, w} norm with `w = 1 / eps`. Currently, for any given sample,
            non-uniform weights are only supported with infinity norm. Example: To specify sample-wise scalar,
            you can provide `eps.shape = (n_samples,) + (1,) * values[0].ndim`.
            
I started debugging the failing tests, but I have trouble finding the issue for the last two tests. It seems that the $L^1$ computation for PGD is different between numpy and pytorch, but I can't figure out why.

Any help would be appreciated.

Cheers!

eliegoudout:
@beat-buesser I think my PR is [ready for review](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382#issue-2072423189)!

eliegoudout:
> Do you think this PR is ready for review (it's currently in draft)?

It is now ready and I would be grateful for any review :) I also fixed the small casting issue that was remaining as well as fixed a small overview.

eliegoudout:
@beat-buesser Hello :) Do you think someone can review / merge this PR at some point?

I'm not in a hurry, but I'm worried it gets forgotten/frozen. Thanks!

beat-buesser:
Hi @eliegoudout Thank you for your patience! I'll review and if possible merge it this week.

beat-buesser:
@eliegoudout Could you please take a look at passing the DCO check?

eliegoudout:
> @eliegoudout Could you please take a look at passing the DCO check?

I'm not entirely sure because I'm not very familiar with this, but I think we're good 👍

beat-buesser:
@eliegoudout Thank you, I think DCO now looks good. Could you please update your branch [eliegoudout:main](https://github.com/eliegoudout/adversarial-robustness-toolbox-fix-FGM-L1/tree/main) with the most recent version of upstream [Trusted-AI:dev_1.18.0](https://github.com/Trusted-AI/adversarial-robustness-toolbox/tree/dev_1.18.0) to minimise the differences to only your changes and fix the merge conflicts?

eliegoudout:
> @eliegoudout Thank you, I think DCO now looks good. Could you please update your branch [eliegoudout:main](https://github.com/eliegoudout/adversarial-robustness-toolbox-fix-FGM-L1/tree/main) with the most recent version of upstream [Trusted-AI:dev_1.18.0](https://github.com/Trusted-AI/adversarial-robustness-toolbox/tree/dev_1.18.0) to minimise the differences to only your changes and fix the merge conflicts?

Again, not proficient with git, but I think it's ok? I simply clicked on "solve merge conflicts" on github, and kept the most recent version in the 3 conflicts it had, is that ok?

beat-buesser:
@eliegoudout Thank you, I think it worked, I'll do a final review next.

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
```diff
@@ -456,54 +457,56 @@ def _compute_tf(
 
     @staticmethod
     def _projection(
-        values: "tf.Tensor", eps: Union[int, float, np.ndarray], norm_p: Union[int, float, str]
-    ) -> "tf.Tensor":
+        values: "torch.Tensor",
```
beat-buesser, line 460:
```suggestion
        values: "tf.Tensor",
```

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
```diff
@@ -456,54 +457,56 @@ def _compute_tf(
...
+        eps: Union[int, float, np.ndarray],
+        norm_p: Union[int, float, str],
+        *,
+        suboptimal: bool = True,
+    ) -> "torch.Tensor":
```
beat-buesser, line 465:
```suggestion
    ) -> "tf.Tensor":
```

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
```diff
@@ -456,54 +457,56 @@ def _compute_tf(
...
-        values_tmp = tf.reshape(values, (values.shape[0], -1))
-
-        if norm_p == 2:
-            if isinstance(eps, np.ndarray):
-                raise NotImplementedError(
-                    "The parameter `eps` of type `np.ndarray` is not supported to use with norm 2."
-                )
+        p = np.inf if norm_p == "inf" else float(norm_p)
+        assert p > 0
+
```
beat-buesser, line 486:
```suggestion

```

art/attacks/evasion/fast_gradient.py
```diff
@@ -431,20 +431,19 @@ def _apply_norm(norm, grad, object_type=False):
...
-                if not object_type:
-                    ind = tuple(range(1, len(x.shape)))
-                else:
-                    ind = None
-                grad = grad / (np.sqrt(np.sum(np.square(grad), axis=ind, keepdims=True)) + tol)
+                i_max = np.argmax(np.abs(grad_2d), axis=1)
+                grad_2d = np.zeros_like(grad_2d)
+                grad_2d[range(len(grad_2d)), i_max] = 1
+            elif norm > 1:
+                q = norm / (norm - 1)
```
beat-buesser, line 442: The variable name `p` and other single character variable names must be replaced with at leaset 3 character names here and all other files to pass the pylint check.

The full pylon report: https://github.com/Trusted-AI/adversarial-robustness-toolbox/actions/runs/8565558666/job/23507533560?pr=2382

art/attacks/evasion/fast_gradient.py
```diff
@@ -334,8 +334,11 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> n
 
     def _check_params(self) -> None:
 
-        if self.norm not in [1, 2, np.inf, "inf"]:
-            raise ValueError('Norm order must be either 1, 2, `np.inf` or "inf".')
+        if not (
+            self.norm == "inf"
+            or self.norm >= 1
```
beat-buesser, line 339: The type checking with `mypy` is complaining here because it does not recognise that `self.norm` can only be `int` on the right side of the `or`. Maybe it could work by translating this statement into an `and` of two `not` instead of using `or` inside `not`?

Full mypy report: https://github.com/Trusted-AI/adversarial-robustness-toolbox/actions/runs/8565558666/job/23507533560?pr=2382

art/attacks/evasion/fast_gradient.py
```diff
@@ -84,7 +84,7 @@ def __init__(
         Create a :class:`.FastGradientMethod` instance.
 
         :param estimator: A trained classifier.
-        :param norm: The norm of the adversarial perturbation. Possible values: "inf", np.inf, 1 or 2.
+        :param norm: The norm of the adversarial perturbation. Possible values: "inf", `np.inf` or a real `p >= 1`.
```
beat-buesser, line 87: Please run `black` formatter on the art folder with `black --line-length 120 art/` with version `black==21.12b0`

beat-buesser (changes requested):
Hi @eliegoudout I have added a first final review for items to pass the style check necessary to be able to merge.

art/attacks/evasion/fast_gradient.py
eliegoudout: I think readability takes a *big* toll because of this, but I used what I think were "reasonable" name variables. In the context of $L^p$ spaces, $p$ and $q$ are quite natural variables, that carry a strong meaning. I think it would have been nice to take advantage of that.

art/attacks/evasion/fast_gradient.py
eliegoudout: Since the problem occurs many times, I prefer creating a float variable
```python
norm: float = np.inf if self.norm == "inf" else float(norm)
```
when necessary.

art/attacks/evasion/fast_gradient.py
eliegoudout: Done. I accidentaly ran black with latest version first, but reran it with `21.12b0` afterwards.

art/attacks/evasion/fast_gradient.py
```diff
@@ -334,10 +336,8 @@ def generate(self, x: np.ndarray, y: Optional[np.ndarray] = None, **kwargs) -> n
 
     def _check_params(self) -> None:
 
-        if not (
-            self.norm == "inf"
-            or self.norm >= 1
-        ):
+        norm: float = np.inf if self.norm == "inf" else float(norm)
```
beat-buesser, line 339: Hi @eliegoudout Did you mean to use `self.norm` instead of `norm`? (twice)

eliegoudout: Yes... 🤦‍♂️

eliegoudout: Fixed it.
It was even 5 times haha. Locally, PGD and FGM tests pass now for me.

art/attacks/evasion/fast_gradient.py
```diff
@@ -334,8 +336,9 @@
 
     def _check_params(self) -> None:
 
-        if self.norm not in [1, 2, np.inf, "inf"]:
-            raise ValueError('Norm order must be either 1, 2, `np.inf` or "inf".')
+        norm: float = np.inf if self.norm == "inf" else float(norm)
```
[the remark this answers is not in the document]
eliegoudout: Fixed in [96a6c93](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382/commits/96a6c93a375c8b7681fddef81aa41eea54e2b36a)

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent.py
```diff
@@ -210,8 +212,9 @@
 
     def _check_params(self) -> None:
 
-        if self.norm not in [1, 2, np.inf, "inf"]:
-            raise ValueError('Norm order must be either 1, 2, `np.inf` or "inf".')
+        norm: float = np.inf if self.norm == "inf" else float(norm)
```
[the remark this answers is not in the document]
eliegoudout: Fixed in [96a6c93](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382/commits/96a6c93a375c8b7681fddef81aa41eea54e2b36a)

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_numpy.py
```diff
@@ -179,8 +181,9 @@
 
     def _check_params(self) -> None:  # pragma: no cover
 
-        if self.norm not in [1, 2, np.inf, "inf"]:
-            raise ValueError('Norm order must be either 1, 2, `np.inf` or "inf".')
+        norm: float = np.inf if self.norm == "inf" else float(norm)
```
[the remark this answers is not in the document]
eliegoudout: Fixed in [96a6c93](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382/commits/96a6c93a375c8b7681fddef81aa41eea54e2b36a)

art/attacks/evasion/fast_gradient.py
beat-buesser: :smiley: Thank you!

eliegoudout:
All jobs pass the tests but fail during `upload to Codecov`. Is this something I can fix or is it only a problem from the runners?

beat-buesser:
@eliegoudout No, but thank you for asking. This is a problem with the codecov service affecting all PRs at the moment. I'm working on a solution.

beat-buesser:
Hi @eliegoudout I have fixed the Codecov issue and now all test are passing, except the one for TensorFlow v1 which looks to fail because of the changes in this PR. Could you please take a look?

eliegoudout:
Okay, very cool, thanks! It seems that these errors occur because I forgot to change the expected value for $x-x_{\text{adv}}$ for BIM:
```
_________________________________ test_images __________________________________
[...]
>           assert np.mean(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.09437845647335052, abs=0.05)
E           assert 0.00080794824 == 0.09437845647335052 ± 5.0e-02
E             comparison failed
E             Obtained: 0.0008079482358880341
E             Expected: 0.09437845647335052 ± 5.0e-02
```
```
_____________________________ test_images_targeted _____________________________
[...]
>           assert np.mean(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.08690829575061798, abs=0.05)
E           assert 0.0004436275 == 0.08690829575061798 ± 5.0e-02
E             comparison failed
E             Obtained: 0.0004436275048647076
E             Expected: 0.08690829575061798 ± 5.0e-02
```
I think it is expected to have a smaller (_i.e._ better) $x-x_{\text{adv}}$. Should I change the tests by simply replacing with the obtained value? Should I keep the same uncertainty? `5.0e-02` seems a bit much. Maybe `abs=1e-4` instead?

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py
```diff
@@ -448,65 +452,60 @@ def _compute_pytorch(
...
+        *,
+        suboptimal: bool = True,
     ) -> "torch.Tensor":
         """
         Project `values` on the L_p norm ball of size `eps`.
 
         :param values: Values to clip.
-        :param eps: Maximum norm allowed.
-        :param norm_p: L_p norm to use for clipping supporting 1, 2, `np.Inf` and "inf".
+        :param eps: If a scalar, the norm of the L_p ball onto which samples are projected. Equivalently in general,
```
beat-buesser, line 467: I think we should add this extended documentation of `eps` also to the class level in line 84 of this files and similar lines for TensorFlow and Numpy implementations.

beat-buesser:
Hi @eliegoudout About the failing test, I think we should take a closer look at why this is happening. `MomentumIterativeMethod` is basically running in that test `ProjectedGradientDescentNumpy` with `norm="inf"`. I think the results for infinity norm should not change because of the existing fast accurate implementation, or?

art/attacks/evasion/fast_gradient.py
```diff
@@ -431,20 +431,19 @@ def _apply_norm(norm, grad, object_type=False):
...
+                grad_2d[range(len(grad_2d)), i_max] = 1
+            elif norm > 1:
+                conjugate = norm / (norm - 1)
+                q_norm = np.linalg.norm(grad_2d, ord=conjugate, axis=1, keepdims=True)
+                with np.errstate(divide="ignore"):
+                    grad_2d = (np.abs(grad_2d) * np.where(q_norm, 1 / q_norm, 0)) ** (conjugate - 1)
+            grad = grad_2d.reshape(grad.shape) * np.sign(grad)
             return grad
 
         # Add momentum
```
beat-buesser, line 449: @eliegoudout I think the failing test is cause by 'norm=1' in line 451 of fast_gradient.py. What do you think, that line should be `grad = _apply_norm(norm=self.norm, grad=grad)`?

beat-buesser: If yes, this was a bug that already existed before this PR and got discovered by your changes.

eliegoudout: Oooh you might be right, nice catch! Indeed, `_apply_norm` should be called with the right `norm` parameter (that is the $p$ of $L^p$, instead of $1$). This also explain perfectly the observed smaller $x-x_{\text{adv}}$, since with this PR, FGM($p=1$) corresponds to only changing the most important input feature (where before, the approach changesd many pixels).

beat-buesser: @eliegoudout Wow, that's great. Thank you for checking the paper, I think your explanation is correct. If not too late I'm wondering if we should add a comment to line 451 to mention that this is a choice by the original paper?

eliegoudout: Yes, I will add a comment, it is a good idea to clarify things I think 👍

eliegoudout: By the way, I think there might be a confusion here (before PR change)
```python
# Add momentum
if decay is not None and momentum is not None:
    grad = _apply_norm(norm=1, grad=grad)
    grad = decay * momentum + grad
    momentum += grad
```
It seems to me that this effectively gives (with $m=\text{momentum}$):
$$m_t = (1+\text{decay})\times m_{t-1} + \text{grad}_t/\Vert\text{grad}_t\Vert_1$$
instead of the paper's

$$m_t = \text{decay}\times m_{t-1} + \text{grad}_t/\Vert\text{grad}_t\Vert_1.$$
Do you agree @beat-buesser?

Furthermore, I think `grad` and `_apply_norm` are poor naming conventions and should be changed. But if we decide to improve them, it should be done later I think.

eliegoudout: > One more note: I think that the design choice of modifying `momentum` only by side effect is pretty confusing. I'll probably leave it like that though, to avoid adding unnecessary refactoring.

Well... as it happens, I just verified, and for reasons I don't know, the in-place modification of momentum doesn't always seem to work. For example, I tried changing the decay factor and got the exact same results: passing tests *both* following cases...
```python
attack = MomentumIterativeMethod(classifier, eps=0.3, eps_step=0.1, decay=1.0, max_iter=10)
x_train_mnist_adv = attack.generate(x=x_train_mnist, y=y_train_mnist)
attack2 = MomentumIterativeMethod(classifier, eps=0.3, eps_step=0.1, decay=123456, max_iter=10)
x_train_mnist_adv2 = attack.generate(x=x_train_mnist, y=y_train_mnist)
assert np.all(x_train_mnist_adv == x_train_mnist_adv2)  # PASSES WHEN IT SHOULD NOT
```

I don't know if this is specific to the tensorflow framework or not, I don't really know how to control which framework pytest uses. In any case, there seems to be a genuine implementation bug regarding momentum.

eliegoudout: Okay, I come with some more understanding of the situation after having implemented my fix regarding the correct computation of the momentum.

For context, the test `test_images` from `test_momentum_iterative_method.py` is essentially as follows:
```python
attack = MomentumIterativeMethod(classifier, eps=0.3, eps_step=0.1, decay=1.0, max_iter=10)
x_train_mnist_adv = attack.generate(x=x_train_mnist)

assert np.mean(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.09437845647335052, abs=0.05)
assert np.max(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.3)
```

and I was really surprised about the error tolerance `abs=0.5` from the first line (>50%!)...

After having updated the code to include my computation of the momentum, I reran the tests with different values for `decay` and frameworks numpy, tf and pytorch. Here are the (surprising) results for the mean of the differences obtained:
```python
# NumPy fixed
{None: 0.09437846,
 0: 0.09437846,
 0.5: 0.1246982,
 1.0: 0.12893921,
 1000: 0.12811291}

# PyTorch fixed
{None: 0.09406391,
 0: 0.09406391,
 0.5: 0.124101475,
 1.0: 0.12861145,
 1000:0.128003241}

# TensorFlow fixed
{None: 0.09437846,
 0: 0.09437846,
 0.5: 0.09437846,
 1.0: 0.09437846,
 1000:0.094378461}
```

My conclusions are as follow:
1. The in-place modification of `momentum` doesn't work, only for TensorFlow (maybe someone knows why?),
2. The person designing the test probably ran them with 2 different framework and concluded, from the size of the difference in results, that it was appropriate to have `abs=0.5`.
3. It seems clear that in fact, the tensorflow version of the momentum iterative method was not working at all (and still isn't!). As I said, I think this might involve refacto, so is left for future PR.
4. In the mean time, we have to statute regarding the tests. I think we should replace the current check with the following, and skip the test for tensorflow atm
```python
# Proposed new check
assert np.mean(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.1288, abs=5e-4)
```

beat-buesser: Hi @eliegoudout That's a great update, let me take a closer look.

beat-buesser: Hi @eliegoudout Thank you very much for your detailed analysis. I agree with your interpretation. The equation for momentum is not implemented correctly. I think removing the wrong inlace addition might also fix the issue with TensorFlow. I like the new test tolerance. We could consider defining it as a relative tolerance, which would make it more interpretable.

beat-buesser: Yes, I agree, let's fix the momentum issue in a different issue and pull request.

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_pytorch.py
eliegoudout: I'm not entirely sure if this is the exact same epsilon. As such, I don't really know what the description should be at the class level to be honest. I think the only difference is that the class-level `eps` is potentially split if the method is iterative, but I'm not sure that there is no other difference. As such, I would appreciate if you could provide the wanted doc for this one.

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
```diff
@@ -342,27 +341,36 @@
...
-            grad = tf.divide(grad, (tf.math.reduce_sum(tf.abs(grad), axis=ind, keepdims=True) + tol))
-            grad = self.decay * momentum + grad
-            # Accumulate the gradient for the next iter
-            momentum += grad
+            raise NotImplementedError(
+                "Momentum Iterative Attack currently not working with the Tensorflow framework. See issue #2439"
+            )
+            # Update momentum in-place (important).
+            # The L1 normalization for accumulation is an arbitrary choice of the paper.
+            grad_2d = tf.reshape(grad, (len(grad), -1))
```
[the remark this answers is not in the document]
eliegoudout: On purpose. Needs rafacto fix.

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
```diff
@@ -342,27 +341,36 @@ def _compute_perturbation(  # pylint: disable=W0221
...
+        # Compute gradient momentum
         if decay is not None and momentum is not None:
-            ind = tuple(range(1, len(x.shape)))
-            grad = tf.divide(grad, (tf.math.reduce_sum(tf.abs(grad), axis=ind, keepdims=True) + tol))
-            grad = self.decay * momentum + grad
-            # Accumulate the gradient for the next iter
-            momentum += grad
+            raise NotImplementedError(
+                "Momentum Iterative Attack currently not working with the Tensorflow framework. See issue #2439"
+            )
```
beat-buesser, line 348: Hi @eliegoudout This exception causes the Pylint error W0101: Unreachable code in the Style Check workflow. I think we can remove the exception because you have opened a separate issue which we will fix for the same release as this pull request.

beat-buesser:
Hi @eliegoudout Could you please take a look at the failing unit tests? It seems the relative tolerance is too tight for some tests.

eliegoudout:
> Hi @eliegoudout Could you please take a look at the failing unit tests? It seems the relative tolerance is too tight for some tests.

Thank you for your message. I do intend to get around this, but I'm struggling to find the necessary time, sorry! I'll do my best :)

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
eliegoudout: So you mean remove "NotImplementedError" line or adding a pylint ignore flag? I think the second approach is much safer personally.

eliegoudout:
> Hi @eliegoudout Could you please take a look at the failing unit tests? It seems the relative tolerance is too tight for some tests.

Done ✅ 
I realized that I didn't properly run both numpy and pytorch frameworks when fixing the test values. I rectified both the expected value and the tolerance (`rel=0.01` when I got approx `0.008` while testing).

art/attacks/evasion/projected_gradient_descent/projected_gradient_descent_tensorflow_v2.py
eliegoudout: I disabled the warning locally as I think it is the safest option atm ([0ddafcd](https://github.com/Trusted-AI/adversarial-robustness-toolbox/pull/2382/commits/0ddafcd83feb70c5a4a91282a7cef4e64e380e91))

beat-buesser:
Hi @eliegoudout Thank you very much! It looks like we need one more change, the MXNet workflow is still failing at two unit tests and requires a larger tolerance for the updated tests. You don't have to install MXNet and can use the information in the test logs to adjust the test tolerances:
```            assert np.mean(x_train_mnist) == pytest.approx(0.12659499049186707, 0.01)
>           assert np.mean(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.1288, 0.003)
E           assert 0.11162121 == 0.1288 ± 3.9e-04
E             comparison failed
E             Obtained: 0.11162120848894119
E             Expected: 0.1288 ± 3.9e-04

tests/attacks/evasion/test_momentum_iterative_method.py:51: AssertionError
_____________________________ test_images_targeted _____________________________

art_warning = <function art_warning.<locals>._art_warning at 0x7f8eb88c4430>
fix_get_mnist_subset = (array([[[[0., 0., 0., ..., 0., 0., 0.],
         [0., 0., 0., ..., 0., 0., 0.],
         [0., 0., 0., ..., 0., 0., 0....0.],
       [0., 0., 0., 0., 0., 0., 0., 0., 0., 1.],
       [1., 0., 0., 0., 0., 0., 0., 0., 0., 0.]], dtype=float32))
image_dl_estimator_for_attack = <function image_dl_estimator_for_attack.<locals>._image_dl_estimator_for_attack at 0x7f8eb88c4f70>

    @pytest.mark.skip_framework("tensorflow")  # See issue #2439
    def test_images_targeted(art_warning, fix_get_mnist_subset, image_dl_estimator_for_attack):
        try:
            (x_train_mnist, y_train_mnist, x_test_mnist, y_test_mnist) = fix_get_mnist_subset
            classifier = image_dl_estimator_for_attack(MomentumIterativeMethod)
    
            attack = MomentumIterativeMethod(classifier, eps=0.3, eps_step=0.1, decay=1.0, max_iter=10)
            x_train_mnist_adv = attack.generate(x=x_train_mnist, y=y_train_mnist)
    
>           assert np.mean(np.abs(x_train_mnist - x_train_mnist_adv)) == pytest.approx(0.1077, 0.01)
E           assert 0.097485796 == 0.1077 ± 1.1e-03
E             comparison failed
E             Obtained: 0.09748579561710358
E             Expected: 0.1077 ± 1.1e-03
```

After fixing this last item, I can merge this pull request. I think your changes are a significant contribution to ART. Therefore, if you like, you may add your employer's name to the AUTHORS file.

eliegoudout:
> Hi @eliegoudout Thank you very much! It looks like we need one more change, the MXNet workflow is still failing at two unit tests and requires a larger tolerance for the updated tests. You don't have to install MXNet and can use the information in the test logs to adjust the test tolerances

Okay, indeed I couldn't test those locally and didn't catch them before, thanks. One question: is this *high* tolerance to be expected? Going from `~1e-3` to `~1e-1` seems really sad, but I'm not at all familiar with this framework so I don't know what to expect, I'll blindly trust you on this.


> After fixing this last item, I can merge this pull request.

Great to read!

> I think your changes are a significant contribution to ART. Therefore, if you like, you may add your employer's name to the AUTHORS file.

Thank you for this recognition, I'm happy this is perceived as significant. Since I made these contributions on my spare time, I would need to discuss this with my company. Let me get back to you on this before the merge (during next week at the latest).

beat-buesser:
Hi @eliegoudout Ok, that sounds great. Instead of one single and large tolerance you could check for the framework and set two expected values with an if-block like this:
```python
def test_generate_parallel(art_warning, fix_get_mnist_subset, image_dl_estimator, framework):
    try:
        classifier, _ = image_dl_estimator(from_logits=True)

        if framework == "tensorflow2":
```
where `framework` contains the label of the backend used for testing. For example:
```python
if framework == "mxnet":
    expected_value = ...
else:
    expected_value = ...
```
Notice to add `framework` as argument to the test function.

eliegoudout:
> Hi @eliegoudout Ok, that sounds great. Instead of one single and large tolerance you could check for the framework and set two expected values with an if-block like this:
> 
> ```python
> def test_generate_parallel(art_warning, fix_get_mnist_subset, image_dl_estimator, framework):
>     try:
>         classifier, _ = image_dl_estimator(from_logits=True)
> 
>         if framework == "tensorflow2":
> ```
> 
> where `framework` contains the label of the backend used for testing. For example:
> 
> ```python
> if framework == "mxnet":
>     expected_value = ...
> else:
>     expected_value = ...
> ```
> 
> Notice to add `framework` as argument to the test function.

@beat-buesser Thanks for clarifying the way to go (I'm not proficient with pytest, would not have thought that the signature would be understood out of the box). I've made the changes.

Regarding the AUTHORS, I wrote my name directly for transparency, as I do not think it have made sense to put my company at this stage.

I hope this is now mergeable :)
Thanks again for all the feedbacks and the help.

Sincerely.
Élie

beat-buesser:
Hi @eliegoudout Thank you very much for your first contribution to ART! We will release ART 1.18 in the coming days which will bring your updates to ART's main branch and packages.

